Issue #6, opened 2017-04-27
Tests assumes English locale


junaruga:
```
____________________________________________ test_run_cmd_exception _____________________________________________

    def test_run_cmd_exception():
        exception = False
        result_e = None
        cmd = 'ls abc'
        try:
            utils.run_cmd_with_capture(cmd)
        except subprocess.CalledProcessError as e:
            exception = True
            result_e = e
    
        assert exception
        assert isinstance(result_e, subprocess.CalledProcessError)
        assert result_e.cmd == cmd
        assert result_e.returncode != 0
        assert result_e.output == b''
        if sys.version_info >= (3, 5):
            assert result_e.stdout == b''
>           assert result_e.stderr == \
                b'ls: cannot access \'abc\': No such file or directory\n'
E           assert b"ls: nelze p...r directory\n" == b"ls: cannot a...r directory\n"
E             At index 4 diff: 110 != 99
E             Left contains more items, first extra item: 101
E             Full diff:
E             - b"ls: nelze p\xc5\x99istoupit k\xc2\xa0'abc': No such file or directory\n"
E             + b"ls: cannot access 'abc': No such file or directory\n"

tests/test_utils.py:56: AssertionError

```

hroncok:
Also note that this is platform dependent, see log from Travis CI:

```
>           assert result_e.stderr == \
                b'ls: cannot access \'abc\': No such file or directory\n'
E           assert b'ls: cannot ...r directory\n' == b"ls: cannot a...r directory\n"
E             At index 18 diff: 97 != 39
E             Right contains more items, first extra item: 121
E             Full diff:
E             - b'ls: cannot access abc: No such file or directory\n'
E             ?  ^                                                  ^
E             + b"ls: cannot access 'abc': No such file or directory\n"
E             ?  ^                  +   +   
```

PR #12, opened 2017-04-29
Enable Travis CI

hroncok:


hroncok:
I've also included a fix of #6 in this PR to make the tests pass.

junaruga:
@hroncok thanks! let me see the code to approve the review.

hroncok:
I got an idea for one slight change.

hroncok:
It worked. Amended. Ready for review.

junaruga:
@hroncok I checked your code. I commented on the code to ask you. Could you check those? Thanks.

hroncok:
I cannot see any comments. Aren't those comments *Pending*?

.travis.yml
```diff
@@ -0,0 +1,28 @@
...
+python:
+- '3.6'
+- '3.5'
+- '3.4'
+
+matrix:
+  include:
+    - python: 3.6
+      sudo: required
+      dist: trusty
```
hroncok: Trusty is a version of Ubuntu. Currently Travis offers "very old Ubuntu" by default and "old Ubuntu" by choosing trusty here. I wen't with trusty as I'm going to do further hacks and was as new ubuntu as possible.

.travis.yml
```diff
@@ -0,0 +1,28 @@
...
+      env:
+      - TOXENV=style
+      before_install:
+      - sudo sed -i 's/trusty/yakkety/' /etc/apt/sources.list
+      - sudo rm -rf /etc/apt/sources.list.d/*
```
hroncok: Line 1: It's changing the Ubuntu release code name in the APT sources list, so when I install `spellcheck`, the version from Yakkety is installed. (Trusty have one version in backports, but that one is old and reports some false errors.)

Line 2: Removing any 3rd party repositories available on Travis CI (not needed, proven to make trouble when doing Line 1).

.travis.yml
```diff
@@ -0,0 +1,28 @@
...
+      - sudo apt-get -qq update && sudo apt-get -qq install shellcheck
+
+cache:
+  directories:
+  - .cache/pip
```
hroncok: Should have been `$HOME/.cache/pip`, sorry about that. Good catch.

hroncok: Amended.

hroncok:
Thanks. I've made it so that Travis CI now has to success before merging.

junaruga:
@hroncok Yes I will follow the way for other PRs. We were able to be ready to sent PRs.

Issue #7, opened 2017-04-27
Not existed app.exit() is not called.


junaruga:
rpmlb/cli.py

```
    app = Application()
    app.run(argv)
    app.exit()
```

`exit()` does not exist in Application class.
But the program is finished at the end of `app.run()` by `raise SystemExit` without passing the `exit()` logic.

rpmlb/app.py

```
    def run(self, argv=None):
...
        raise SystemExit(is_error)
```

This is not good.

Remove `raise SystemExit`, and use below logic in `cli.py`.

```
sys.exit(app.run(argv))
```

PR #14, opened 2017-05-03
Fix the issue for returning exit status.

junaruga:
This fixes https://github.com/sclorg/rpm-list-builder/issues/7

Let me explain about below added part of `rpmlb/cli.py`

```
+    is_success = app.run(argv)
+    sys.exit(0 if is_success else 1)
```

The reason why I use "is_success" variable is because I prefer "1 statement per line" policy [1].
I also prefer the way of that only `cli.py` knows, and manages the concept of "exit status", and  `app.py`'s Application class does not know it in terms of objective oriented.

http://python-guide-pt-br.readthedocs.io/en/latest/writing/style/#one-statement-per-line


rpmlb/app.py
```diff
@@ -21,16 +21,18 @@ def __init__(self):
 
     def run(self, argv=None):
         LOG.info("Starting %s (%s)", self._program, self._version)
-        is_error = False
+        is_success = False
         try:
             self.run_internally(argv)
+            is_success = True
```
hroncok: Move this to the `else` part of the `try-except-else` statement. You are only trying the thing above.

tests/test_app.py
```diff
@@ -8,6 +9,20 @@ def test_init():
     assert app
 
 
+def test_run():
```
hroncok: Test names have to be self-explanatory. What about `test_run_calls_run_internally`?

hroncok: Oh, now I see you are actually testing 2 things here. The return value and the fact that this was run. Please split it to 2 tests. One test should assure that `run_internally` was being called. The other one should  care for the exit code and can use a lambda statement instead of mock...

tests/test_app.py
```diff
@@ -8,6 +9,20 @@ def test_init():
...
+    app = Application()
+    app.run_internally = mock.MagicMock(return_value=True)
+    assert app.run()
+    assert app.run_internally.called
+
+
+def test_run_with_exception():
```
hroncok: And in this test, just check the return value. I suggest 3 following tests:

 * `test_run_calls_run_internally` (use mock)
 * `test_run_returns_true_on_success` (use lambda)
 * `test_run_returns_false_on_exception` (use lambda)

hroncok:
Also, please if a commit fixes a GitHub issue, include this in the commit message:

```
Fixes: https://github.com/sclorg/rpm-list-builder/issues/7
```

rpmlb/app.py
junaruga: `try-except-else`! OK I try to use it.

junaruga:
> Fixes: https://github.com/sclorg/rpm-list-builder/issues/7

Below is better, isn't it?
Because it's more simple.
And when we will change the repository as a possibility in the GitHub, it still works.

```
Fixes #7
```

tests/test_app.py
junaruga: OK. I will keep the rule of yours: "one test method, one test topic".

tests/test_app.py
junaruga: OK I will do it!

junaruga: I can not imagine how to use `lambda` for above tests.
Do I use `lambda` to replace `run_internally`'s logic?

hroncok:
>     Fixes #7

That's not better. because if we decide to go from github to another service, the link still makes sense. Also, GitHub makes the long link appear as just  #7, so it's looking the same but holds more information.

tests/test_app.py
hroncok: Oh, you cannot raise an exception from lambda (I forgot). In that case, you can do:

```python
def run_internally(*args):
    raise RuntimeError(...)
app.run_internally = run_internally
```

junaruga:
OK. I agree with you.
If we consider the case to move the repository, only `#number` is confusing us which git service's issue number it is. Your way is better for the case.
I will follow your way.

tests/test_app.py
junaruga: Why is your above way, defining run_internally manually for the test is than using mock?
Using mock looks better for  me.

```
app.run_internally = mock.Mock(side_effect=RuntimeError(...))
```

hroncok: Mock is "magical". It's good idea to avoid it when possible and only use it if no better way exists.

junaruga: Sure. It makes sense.

junaruga:
@hroncok I fixed all the items that you mentioned, amended, and the test was passed.

junaruga:
Thanks for the reviewing and merging.

PR #17, opened 2017-05-04
Dynamic imports of downloaders and builders

khardix:
Judging from the `# TODO: Use reflection.` comments, the current mechanism for loading downloaders and builders was a temporary solution.

I went ahead and tweaked the importing functions in `.utils`, so they now properly provide the requested class, and switched the `Base{Builder,Downloader}.get_instance` methods to utilize them.

As a bonus, I wrote [Napoleon](http://www.sphinx-doc.org/en/1.5.5/ext/napoleon.html)-style docstrings for all affected functions/methods.


rpmlb/builder/base.py
```diff
@@ -5,6 +5,8 @@
 
 import retrying
 
+from .. import utils
+
 # from rpmlb import utils
```
hroncok: Maybe also remove that in  the process?

rpmlb/builder/base.py
```diff
@@ -16,29 +18,23 @@ class BaseBuilder:
...
-            from rpmlb.builder.dummy import DummyBuilder
-            instance = DummyBuilder()
-        else:
-            raise ValueError('name is invalid.')
+    @staticmethod
+    def get_instance(name):
+        """Instantiate named builder.
+
+        Keyword arguments:
+            name (str): The name of the requested builder.
```
hroncok: If you already know this takes str, maybe use type annotation in function definition instead?

rpmlb/builder/base.py
khardix: Ah, missed that comment :) I will remove it.

rpmlb/builder/base.py
khardix: Good point, I forgot we are not on Py2-compat mode any more 🍾 .

PR #19, opened 2017-05-09
Change project URL in setup.py.

junaruga:
Change our project URL in `setup.py`


junaruga:
@omron93 thanks for the review!

PR #21, opened 2017-05-17
CLI interface using click

khardix:
As suggested on the initial review meeting, I went ahead and rewrote
the CLI interface to use [click][].

[click]: http://click.pocoo.org/6/ "A command line library for Python."

I attempted to port it as directly as possible. The most significant changes
are:
-   Drop of the `app` module (and subsequently the `Application` class). Their
    responsibilities were taken by the `cli` module and its `run` function.
    The `Application.run` logic was dropped altogether -- click itself handles
    keyboard interrupts and setting of appropriate exit codes.
-   Most of the application tests (`tests/test_app.py`) were dropped as well --
    I see no reason why we should test that the click does its job. The only
    test ported was `test_parse_argv_no_options`, both as a way to ensure
    expected default values and as an example of testing a click application.
-   Adjusted entry point in `setup.py` and add `__main__` module. This enables
    the package to be run directly as a module (`python -m rpmlb`) as well as
    console script.


khardix:
Well, tests fail on missing click (even after I added it into the `requirements.txt`) – what I am missing?

hroncok:
That's unwanted side effect of caching `.tox`. After purging the cache, the build passes. Locally, removing `.tox` before running it workarounds that issue. Maybe there is a tox option that installs the dependencies again? Not sure.

junaruga:
`tests/test_app.py` is a test file for `rpmlb/cli.py`?
It does not make sense for me.

I prefer using `rpmlb/app.py` instead of `rpmlb/cli.py`, removing `rpmlb/cli.py`.
Because my image for Application is that Application manages actual top level's logic in the application.
CLI means Command Line Interface. It is just interface. It does not have an actual content.

It's like `cli.py` and `application.py` file in `falke8`.
https://github.com/PyCQA/flake8/tree/master/src/flake8/main

junaruga:
> Adjusted entry point in setup.py and add __main__ module. This enables
the package to be run directly as a module (python -m rpmlb) as well as
console script.

Running command line as `python -m rpmlb` is popular way?
`bin/rpmlb` type's command has not been available any more?

junaruga:
@khardix I think that this PR is too big. It includes several topics that need discussion.

I think that a contribution is like a "Tetris" or "Chess".
Effective acceptable PR is proceeding your pull-request step by step observing the others' reaction and needs.

For example, if your final expected result is filling below 3 points,

- Replacing `argparse` to `click`. => Feature A
- Moving main logic `app.py` to `cli.py`, removing `app.py`. => Feature B
- Changing entry point interface (Adding `python -m rpmlb`) => Feature C

You can think your plan as

- Send PR A for feature A (=> like Tetris Block A)
- If PR A is accepted, send PR B for feature B (=> like Tetris Block B)
- If PR B is accepted, send PR C for feature C (=> like Tetris Block C)

```
PR A => PR B => PR C
```

or like a Chess,

```
PR A +-> If Reaction 1 -> PR B -> If Reaction 3 -> PR C
     |                         -> If Reaction 4 -> PR C'
     +-> If Reaction 2 -> PR B'
```

Planning PRs, you can start the PR from easily accepted feature.
This is like a Sales person's "Yes set close" strategy.

Right now I am positive for the feature A. But still I want to ask you why and what the benefit is.

And I am negative for the feature B and C.
I think that those B and C makes `rpmlb` worse. I have a clear reason for that.

We may need discussion for B and C on issue pages.

But at first I want another PR for only feature A.

I can understand that you have very detailed knowledge about Python.
But I guess that we have different perspective about a topic of objective oriented design, secure programming design and usability for each other. It is not programming language specified topic. It's more general topic.
So, I think we can recognize this fact.

hroncok:
> - Replacing `argparse` to `click`. => Feature A

This is definitely desired.

> - Moving main logic `app.py` to `cli.py`, removing `app.py`. => Feature B

Let em have a look at that one.

> - Changing entry point interface (Adding `python -m rpmlb`) => Feature C

That one is very pythonic and I like that.

junaruga:
@hroncok do you like that this PR title (= CLI interface using click = Feature A) is actually including feature A, B and C? And in this case, I guess that feature B and C are not necessary features to achieve the feature A.
In other projects' cases. When I sent PR including not directly related features by my mistake, I have declined by the project maintainer. One PR one feature is better, isn't it?

hroncok:
Here it's highly related. I don't consider it excellent practice, but I don't feel so against it.

khardix:
For start, let me state that the discussion here is exactly what I was hoping for – recognizing different patterns of development and agreeing on how to do things :)

## TL;DR

1. Yes, it is a big chunk, but doing it in parts would require lot of extra (and in the end pointless) work.
2. The entry point is not changed, but added – you should be able to use both `${prefix}/bin/rpmlb` and `python -m rpmlb` now with the same results.

## Long version:

> - Replacing `argparse` to `click`. => Feature A

I agree it is a big PR. However, `argparse` and `click` have different philosophies on how to specify the parameters. Mapping `click` way to existing `Application` class would be very unwieldy. What is more, after removing the parts that are handled by `click` itself, I would end with class with only one method – which is considered an anti-pattern ([talk][stop-writing], [followup][oo-heresy]). As a personal opinion, I also dislike classes that will only ever have one instance during a run. In the end I went and transformed the `Application` class into a single function (`cli.run`).

[stop-writing]: http://pyvideo.org/pycon-us-2012/stop-writing-classes.html "Stop Writing Classes"
[oo-heresy]: https://eev.ee/blog/2013/03/03/the-controller-pattern-is-awful-and-other-oo-heresy/ "The controller pattern is awful (and other OO heresy)"

In conclusion, I would also like smaller PR, but in this case it would mean an unreasonable amount of extra work.

> - Moving main logic `app.py` to `cli.py`, removing `app.py`. => Feature B

I really do not care if the module will be called `app` or `cli`, but I tend to use `cli` for "module holding functions defining the command-line interface". Also the name of the function `run` is really generic, but I did not think of better one yet.

> - Changing entry point interface (Adding `python -m rpmlb`) => Feature C

It is not changing per se – the `${prefix}/bin/rpmlb` is still generated on `pip install` or equivalent (thanks to `setup.py` `entry-points` parameter). I merely added the option to run it without installation (provided that all dependencies are in place, of course) with `python -m`.

junaruga:
> The entry point is not changed, but added – you should be able to use both ${prefix}/bin/rpmlb and python -m rpmlb now with the same results.

Sure. That's good news. ok I agree adding `__main__.py`.

I am worry that after your change, the test coverage rate becomes lower.
Do you know the concept of "test coverage"?
If the code is not used by the unit test, how do we prove the code is correct.
Manually? No. It does not make sense.

I found the unit test for the code with `click`.
http://stackoverflow.com/questions/26767104/testing-click-python-applications

Moving the logic from `app.py` to `cli.py`.
OK. Please go a head.

The merit of keeping very small `cli.py` as much as possible, was that we could test top level logic with creating Application.class with `py.test` with mocking.
But the testing is good, it does not matter.

I think what you need is the experience to contribute to other popular python software.
As I know you in person, I would explain detail, and won't close this PR from myself.

But if you contribute your code and send PR to other packages maintained by unknown people, people say "Sorry your code does not make sense for me. Close this ticket." Just like this.

What you need is the experience that you are declined or failed something, or what you do what you feel reasonable. Then you will see some common senses or a difference of making a difference in your experience.

junaruga:
By the way, now you have 8 commits in this PR, so finally you will rebase those to 1 commit, right?

hroncok:
> By the way, now you have 8 commits in this PR, so finally you will do rebase and squash from those commits to 1 commit, right?

Not necessarily. In this case, I'd say no.

junaruga:
> Not necessarily. In this case, I'd say no.

OK you like it.
I know that some people like 1 PR 1 commit in my experience.
But it does not matter.

junaruga:
> Not necessarily. In this case, I'd say no

@hroncok I want to ask you which kind of case is yes, and which kind of case is no for you?

hroncok:
>  I want to ask you which kind of case is yes, and which kind of case is no for you?

If the commits bare significant logic by being separate I deem that appropriate. Example:

 1. Introduce a new API for better beer experience
 2. Drink all beers using the new API, because it tastes better
 3. Add several new beers to our collection (because the new API makes that possible)

If they don't, I tent to squash. Example:

 1. Implement new Feature
 2. Add test for Feature
 3. Fix typo in Feature
 4. Changes of Feature based on review

junaruga:
@hroncok  beer? you mean deer 🍺 ? Or this is not a implicit or metaphor to another meaning, is it?
All right now I knew your needs and the way how you achieve it.

I agree with your needs and the result that you want to gain.
Though maybe we have a different opinion about way to achieve it for each other, it does not matter.
Thanks for your explanation.

hroncok:
I mean beer as in :beer: No metaphor, just an example.

khardix:
> I am worry that after your change, the test coverage rate becomes lower.

Valid point. I will go through the options and try to come up with reasonable tests. Suggestions for what to test for (i.e. "Ensure that all paths are absolute") are welcome.

> By the way, now you have 8 commits in this PR, so finally you will do rebase and squash from those commits to 1 commit, right?

I agree with @hroncok here – I feel that squashing the PR to one commit would needlessly throw away the git history.

khardix:
So, I have pushed few example tests for the CLI. As I have only vague idea what to test for, I would like get early feedback and save us all some time. Any nitpicking appreciated ☺.

junaruga:
I have not seen your code yet, as I am PTO until this Friday.

At least fill test cases of mine that you deleted. If possible, try 100 percent coverage about the code that you updated.
Another PR for pytest cov helps that.
You can check your code actual line of the code is covered by the html report detail page.

junaruga:
About this PR, I will delegate the review to Miro.
I won't review it.

tests/test_cli.py
```diff
@@ -0,0 +1,133 @@
...
+    assert LOG.getEffectiveLevel() == logging.INFO
+
+    recipe = Path('recipe.yml')
+    recipe.touch()
+
+    runner.invoke(run, ['--verbose', str(recipe), 'test'])
+    assert LOG.getEffectiveLevel() == logging.DEBUG
+
+    runner.invoke(run, [str(recipe), 'test'])
+    assert LOG.getEffectiveLevel() == logging.INFO
```
hroncok: Could you split this to a parametric test with verbose set tot True and False? Testing just one thing in one test is better. (Let me know if you need help with that.)

tests/test_cli.py
```diff
@@ -0,0 +1,133 @@
...
+    # Nonexistent parameter
+    with pytest.raises(click.BadParameter):
+        run.make_context('test-work_dir-nonexistent', options + arguments)
+
+    # Fine parameters
+    work_dir.mkdir()
+    ctx = run.make_context('test-work_dir-ok', options + arguments)
+    result = Path(ctx.params['work_directory'])
+    assert result.is_absolute(), 'Relative work directory path'
+    assert result == work_dir, 'Unexpected work directory'
```
hroncok: Again here. Plese split to two tests. Either with parametrization, with fixtures or with common function for the setup part.

tests/test_cli.py
```diff
@@ -0,0 +1,133 @@
...
+    custom_file.touch(mode=0o200)
+    with pytest.raises(click.BadParameter):
+        run.make_context('test-custom_file-unreadable', options + arguments)
+
+    # Proper result
+    custom_file.chmod(0o600)
+    ctx = run.make_context('test-custom_file-ok', options + arguments)
+    result = Path(ctx.params['custom_file'])
+    assert result.is_absolute(), 'Relative custom file path'
+    assert result == custom_file, 'Unexpected custom file path'
```
hroncok: And here as well. One test tests one thing.

khardix:
Looking into [pytest docs on parametrizing](https://docs.pytest.org/en/latest/fixture.html#parametrizing-fixtures), I can see how it may be done, but it did not give me a clear idea on usage (i.e. how to determine expected result from the parametrized value, apart from the obvious big if-statement (or equivalent) with branch for each possible value).

Any best-practice examples of the usage of the parametrization?

hroncok:
> Any best-practice examples of the usage of the parametrization?

Nothing that I know of. Let me send you a proposal of how would i do it?

hroncok:
See https://github.com/khardix/rpm-list-builder/pull/1

khardix:
Thanks. I merged the changes and will try to write the rest of the tests more like that.

khardix:
After the last commit, all the test I could think of should be included. If there will be no further issues with them, I will squash the test-related commits into one (which should also fix the hung-up Travis check, hopefully).

@hroncok, can I ask for re-review (at your convenience)?

hroncok:
I will re-review once rebased to solve the merge conflicts, as discussed on IRC.

Issue #20, opened 2017-05-09
cmd element in recipe file


junaruga:
I suggested below proposal.

This ticket manages the actual implementation for the proposal.
https://github.com/sclorg/rhscl-rebuild-recipes/issues/17

junaruga:
Memo: Maybe I can start to implement for this feature after #32, #33 will be done, because this feature is deeply related to those tasks, and the `cmd` is run in the same timing with that the RPM spec file is edited by `macros`, and `replaced_macros`.

PR #30, opened 2017-05-24
Custom file hacks necessary for rh-ror50

khardix:
As said in title, this PR contains the shell hacks necessary for successful rebuilding of the `rh-ror50` SCL.
After the [recipe fixes](https://github.com/sclorg/rhscl-rebuild-recipes/pull/18) merge, it should be possible to rebuild the collection from scratch.

Of course, these hacks should be removed after they will be included in respective spec files.

**Note**: I'm not sure about the location and name of the file (`tests/fixtures/custom/rhpkg_internal_copr.yml`) – right now it feels as this file is part of the automated tests. Is that intentional, or should we rename/move the file to more appropriate location?

### DIFF
--- a/tests/fixtures/custom/rhpkg_internal_copr.yml
+++ b/tests/fixtures/custom/rhpkg_internal_copr.yml
@@ -18,6 +18,15 @@ build:
         sed -i 's/rh-nodejs6/rh-nodejs4/g' rh-ror50.spec
         sed -i 's/rh-mongodb32/rh-mongodb26/g' rh-ror50.spec
     fi
+  - | # Hack around flexmock bootstrap
+    if [ "$(basename $(realpath ..))" = "009" ]; then
+        sed -i '/^BuildRequires.*rubygem(rspec)$/ s/^/#/' rubygem-flexmock.spec
+        sed -i '/^%check$/,/^popd$/ s/^/#/' rubygem-flexmock.spec
+    fi
+  - | # Hack around railties dependency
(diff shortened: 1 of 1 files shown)


hroncok:
> right now it feels as this file is part of the automated tests

What is this actually used for?

khardix:
> What is this actually used for?

I was pointed to that as a custom file for building the `rh-ror50` collection. It contained some `ror50`-related hacks in the beginning, so I went with it and with @junaruga's help expanded it to built the whole collection.

Other than that, it is mentioned in [the example how-to](https://github.com/sclorg/rpm-list-builder/wiki/Example-to-build-with-Copr-for-Ruby-on-Rails-recipe) on wiki here.

hroncok:
I'm not sure rpm-list-builder should store files specific to certain SCL collections (except for testing).

junaruga:
I am fine for this modification right now.
Ideally those may be managed in the recipe file.
But right now there is no way for that.

So, after below implementation, we can remove the those hack from this file again.
https://github.com/sclorg/rpm-list-builder/issues/20

But right now I am fine for this modification.
Thanks.

The file can be used as kind of "valid" sample file for rh-ror50 SCL.


### For detail of the modification.

* `rubygem-flexmock`'s hack is needed because of the bug of the RPM spec file, as right now we can not push the fixed version to the repo.
* `rubygem-railties`' hack is for only CentOS 7. Copr is using CentOS7 right now. This hack is not needed for Fedora and RHEL. Umm, this hack also ideally can be managed in the recipe file with flatform condition flag.

PR #34, opened 2017-06-02
Change "properties without logic" to "instance variables".

### LINKED ISSUE #23: Rafactoring code
From @hroncok code review.

```
genreally:
  sometimes, docstrings would help
    often missing or very non verbose
    often captain obvious
  properties without logic? (recipe, work)
    trying to make it public read only? "consenting adults"

 __init__.py:
  nothing belongs here

app.py
  some weird commented out things?
  the Namespace class is very non-pythonic, it's "magical"

custom.py
  naming things is hard, but Custom?

recipe.py
  verify is bad.
    assserts for logic
    create a verify_attribute function?

utils.py:
  the os.environ manip is unsafe (fixed)
  dict update? (fixed)
  unly *try*  things that are needed, not entire wall of code

work.py:
    mixing format and %

yaml.py:
    import not at the begining of file without reason
    overobjectified

builder:
  base:
    you don't need empty inits
      and if you do, raise NotImplementedError or use ABC
      https://docs.python.org/3/library/abc.html
    explain empty after and before method in docstrings

I'd run integration tests using pytest as well.
I'd also drop the Makefile. Let the developer decide how to create the venvs.
```

junaruga:
This fixes #23 only below part.

>   properties without logic? (recipe, work)
>    trying to make it public read only? "consenting adults"

I found out that `verify` method was not called.
So, I added 1 line to call verify method at `app.py`.
Ideally I should add the unit test to `test_app.py` for that.
But I did not add unit test for that to `test_app.py` this tme.
Because it's to prevent conflict. PR #21 is updating the file. 

Instead of that, I checked that `$ make venv-integration-test` was passed.

### DIFF
--- a/rpmlb/app.py
+++ b/rpmlb/app.py
@@ -43,6 +43,8 @@ class Application:
 
             # Load recipe
             recipe = Recipe(args_dict['recipe_file'], args_dict['recipe_id'])
+            # Verify recipe data
+            recipe.verify()
 
             work = Work(recipe, **args_dict)
             # Load downloader
--- a/rpmlb/recipe.py
+++ b/rpmlb/recipe.py
@@ -15,24 +15,15 @@ class Recipe:
             raise ValueError('recipe_id is required.')
 
         self._recipe_id = recipe_id
-        self._recipe = None
 
         yaml = Yaml(file_path)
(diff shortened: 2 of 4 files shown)


junaruga:
Thanks for the checking. Merged.

Issue #18, opened 2017-05-05
Builder to modify modified RPM spec file when using --resume


junaruga:
## Situation

For example, a build failed at position 5.

```
$ rpmlb \
  --build custom \
  --custom-file tests/fixtures/custom/rhpkg_internal_copr.yml \
  --work-directory $HOME/tmp/rpmlb_work_dir \
  ../rhscl-rebuild-recipes/ror.yml \
  rh-ror50 2>&1 | tee rpmlb.log
```

When it is run from position with resume option.

```
$ rpmlb \
  --build custom \
  --custom-file tests/fixtures/custom/rhpkg_internal_copr.yml \
  --work-directory $HOME/tmp/rpmlb_work_dir \
  --resume 5 \
  ../rhscl-rebuild-recipes/ror.yml \
  rh-ror50 2>&1 | tee rpmlb.log
```

```
$ cd tmp/rpmlb_work_dir/005/rubygem-diff-lcs
```

Updating to add `%global _with_bootstrap 1` to the RPM spec file that has already been updated on last build.

```
$ git diff rubygem-diff-lcs.spec
+# Edited by rpmlb
+%global _with_bootstrap 1 <== update by 2nd build.
+# Edited by rpmlb
+%global _with_bootstrap 1 <== update by 1st build.
+
...
```

This situation does not become an error as it is technically valid.
It does not become surface .

## Possible Solution

`https://github.com/sclorg/rpm-list-builder/blob/master/rpmlb/builder/base.py prepare function

1. will read the RPM spec file data from `foo.spec.orig` file if it exists.
2. if the orig file does not exist, will read from `foo.spec` file to get the RPM spec file data.

We can not use `git reset --hard HEAD` or other git command to reset the file content internally in `rpmlb` program because we will consider non-git environment too.

khardix:
I did a typo in [commit message that should closed this one](https://github.com/sclorg/rpm-list-builder/commit/32a4a46c2e63f016f931b2ad7d7987c8afd70157). Closing manually.

Issue #33, opened 2017-05-30
[RFE] Modularize the SPEC file editing


khardix:
## The problem

There are several ways we need to edit the SPEC file during the build. The current solution has [issue](https://github.com/sclorg/rpm-list-builder/issues/18)(s), and is not very flexible. Since a builder might want to add specific edits not needed otherwise (i.e. some macros need adjustment only when building in koji), the system needs to change.

## Proposed solution

Let's leverage the python's excellent iteration protocol and generators:

1.  Have one function (context manager) that prepares and finalizes the editation:
    -   Backs up the original file if necessary.
    -   Opens the original for reading and the destination file for writing.
2.  Make the editations into generators, which generate lines of the result file.

### Possible implementation overview

```python
@contextlib.contextmanager
def edit_spec_file(self, spec: pathlib.Path):
    """Safely edit a spec file.

    Keyword arguments:
        spec: The spec-file to edit.

    Yields:
        Open handles for input and output file.
    """

    # Back-up the file & other preparation

    # Both open files are safely closed afterwards
    with original.open(mode='r') as src, target.open(mode='w') as dst:
        yield src, dst


def add_macros(self, source: Iterator[str], macros: Mapping[str, str]):
    """Add macros to the source stream."""

    for macro, value in macros.items():
        yield '%global {macro} {value}'.format(macro=macro, value=value)

    yield from source


def edit_macros(self, source: Iterator[str], macros: Mapping[str, str]):
    """Edit macros already present in the file."""

    for line in source:
        if matches_macro_regex(line):
            yield edited_line
        else:
            yield line

def extra_edits(self, source: Iterator[str]):
    """Builder-specifif SPEC file edits. Override if needed."""

    yield from source  # pass for generators
```

Usage:

```python
with self.edit_spec_file(path_to_spec) as (src, dst):
    edited = iter(src)

    edited = self.add_macros(edited, macros_to_add)
    edited = self.edit_macros(edited, macros_to_edit)
    edited = self.extra_edits(edited)
    # add any other file processing if needed

    dst.write('\n'.join(edited))
```

*Note:* For details on how it works, see [David Beazley's tutorial on generators](http://www.dabeaz.com/generators/).

junaruga:
It does not matter.

khardix:
@junaruga I'm not sure what you meant by the response – do you mean that you do not care which way it would be implemented?
@hroncok Can I ask about your thoughts on the matter?

junaruga:
[...]

PR #35, opened 2017-06-13
Fix double spec modification

khardix:
In preparation for implementation of #33, I rewrote the `BaseBuilder.edit_spec_file` function to a context manager. This change also adds a test for and fixes issue #18.

### DIFF
--- a/rpmlb/builder/base.py
+++ b/rpmlb/builder/base.py
@@ -2,6 +2,7 @@ import logging
 import os
 import re
 import sys
+from contextlib import contextmanager
 
 import retrying
 
@@ -94,7 +95,9 @@ class BaseBuilder:
     def build(self, package_dict, **kwargs):
         raise NotImplementedError('Implement this method.')
 
-    def edit_spec_file(self, spec_file):
+    @staticmethod
+    @contextmanager
+    def edit_spec_file(spec_file):
         spec_file_origin = '{0}.orig'.format(spec_file)
         os.rename(spec_file, spec_file_origin)
         fh_r = None
@@ -114,7 +117,7 @@ class BaseBuilder:
         if not isinstance(macros_dict, dict):
             return ValueError('macros should be dict object.')
 
-        for fh_r, fh_w in self.edit_spec_file(spec_file):
+        with self.edit_spec_file(spec_file) as (fh_r, fh_w):
             for key in list(macros_dict.keys()):
                 value = macros_dict[key]
                 if value is None or str(value) == '':
@@ -128,7 +131,7 @@ class BaseBuilder:
         if not isinstance(macros_dict, dict):
             return ValueError('macros should be dict object.')
 
-        for fh_r, fh_w in self.edit_spec_file(spec_file):
+        with self.edit_spec_file(spec_file) as (fh_r, fh_w):
             for line in fh_r:
                 line = line.rstrip()
                 for key in list(macros_dict.keys()):
--- a/rpmlb/builder/base.py
+++ b/rpmlb/builder/base.py
@@ -1,8 +1,8 @@
 import logging
-import os
 import re
 import sys
 from contextlib import contextmanager
+from pathlib import Path
 
 import retrying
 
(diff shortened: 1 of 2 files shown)


junaruga:
@khardix I saw that I am assigned as a reviewer in this PR.
Sorry I do not want to review your Python code. I am not a right person for that.
Please ask @hroncok for review.

khardix:
@junaruga As you wish.

rpmlb/builder/base.py
```diff
@@ -94,27 +95,40 @@ def build_with_retrying(self, package_dict, **kwargs):
...
+        # Ensure path type
+        target = Path(target) if not isinstance(target, Path) else target
+
+        # Back up the original
+        source = target.with_suffix('.spec.orig')
+        if not source.exists():
+            target.rename(source)
+
+        # Provide the handles
+        with source.open(mode='r') as src, target.open(mode='w') as tgt:
```
junaruga: After talking with mhroncok on IRC, let me comment. Other things look good.

## 1.
Variable `source` variable is an object of `Path` class. Variable `src` variable is an object of `File` class(?) And `src` is shorten word of `source`.
Variable `target` is an object of `Path` class. Variable `tgt` is an object of `File` class? And `tgt` is shorten word of `target`.

Using same concept (meaning)'s word for different class is usual in Python world?
It is not usual for me in terms of naming consistency. (1 kind of naming for 1 concept).

`target` as `Path` object, `target_file` as `File(?)` object looks better isn't it?

Below sample using `p` as `Path`, `f` as `File(?)` as reference.
https://docs.python.org/3/library/pathlib.html#pathlib.Path.open

## 2.
About using shorten word, which case is popular for Python world?

a. Use not shorten word as variable as much as possible such as `target`. Use shorten word for known word as exception `tgt`.
b. Use shorten word as variable usually.  

My understanding is for example,
Perl (Modern), Ruby, Java, C#, JavaScript, PHP are case a.
C, Perl (Traditional) are case b.

junaruga:
About naming convention "The Art of Readable Code" was useful.
https://www.amazon.com/dp/B0064CZ1XE

rpmlb/builder/base.py
hroncok: 2. In Python, you usually only shorten things in generator expressions etc.

hroncok: 1. Since here we are not exposing the names any further, I'm OK with both ways (e.g. changing it to `target_file` or keeping it as is).

junaruga: 2. => OK, thanks for the information. I saw `flask`'s source code now. `flask` is one of my compass source code :)

It looks that not only some known words such as `app`, `args`, `fp`, but also other words `exc`, `ctx` I could not recognize :) are shorten. The rate is a little bit higher than Ruby and Perl in my impression.

khardix: Well, the main reason for "shortcuts" such as `exc`, `src`, `tgt`, etc. is the attempt to squeeze some expression/statement to one line – here the double file opening. With longer names, `flake8` forces me to do the following, which seems ugly to me:

```python
with source.open(mode='r') as source_file, \
        target.open(mode='w') as target_file:
```

Maybe we can discuss the 80-character-per-line limit in separate issue – personally, I have it set to 100 and have far less issues as this.

But here I will go with lesser evil and extend the names to their proper form.

khardix:
@junaruga The names should be better now.

tests/builder/test_base.py
```diff
@@ -1,9 +1,25 @@
...
 from unittest import mock
 
+import pytest
+
 from rpmlb.builder.base import BaseBuilder
 
 
+@pytest.fixture
+def empty_spec_path(tmpdir):
+    """Empty spec file path"""
```
junaruga: Sorry one more thing.
Did you see below below `pushd_tmp_dir` function? It is same kind of function with `empty_spec_path`, isn't it?
Why do you create same kind of method as different file and different implementation?
Is it possible to implement it on DRY (Don't repeat yourself) concept?

Also variable `tmp_dir` is better than `tmpdir`?
https://github.com/sclorg/rpm-list-builder/blob/master/tests/helper.py#L23

junaruga:
> @junaruga The names should be better now.

OK, it makes sense for me. Thanks.

rpmlb/builder/base.py
junaruga: > Maybe we can discuss the 80-character-per-line limit in separate issue – personally, I have it set to 100 and have far less issues as this.

OK, please create new ticket with the reason why you prefer 100 than 80 to discuss that.
As a information, `flake8`'s default max length for a line is 79.

tests/builder/test_base.py
khardix: `tmpdir` is a built-in fixture provided by the `pytest` framework itself. I did not see the `push_tmp_dir` function, since I did not look for it and just used the built-in one.

`empty_spec_path` ensures freshly created empty spec file for each test case that would need it. I agree that the pattern is similar to `push_tmp_dir`, but I'm not aware on how to implement that generically enough to be usable for both files and dictionaries at the same time.

junaruga:
Could you please rebase this PR because the Travis test bug was fixed?

khardix:
> Could you please rebase this PR because the Travis test bug was fixed?
@junaruga Done.

PR #37, opened 2017-06-20
Modular spec editation

khardix:
This is a proposed implementation of sclorg#33 – see discussion there for details.

As a forewarning, this change tuned to be a bit larger than expected. If lost, try to inspect one commit at a time, which should show you how the change evolved.

What is implemented:
- The SPEC file preparation turned into a sequence of generators. This should isolate atomic changes and ease further modification.
- Base builder provide a method that can be overridden in derived classes, which allows them to hook into the preparation process.
- Unit tests were added for all modified or added functionality.


rpmlb/builder/base.py
```diff
@@ -11,6 +12,17 @@
...
 
+#: Regular expression for finding macro definitions
+MACRO_REGEX = re.compile(
+    r'''^
+    %global\s+                   # beginning of definition
+    (?P<name>[^\s]+)\s+          # one-word name
+    (?P<value>(?:.|(?<=\\)\n)+)  # value including spaces and escaped newlines
+    $''',
+    flags=re.MULTILINE | re.VERBOSE,
+)
```
khardix: I'm not sure about the regex definition on module level. Currently it is needed only by a single method, an I'm getting a bit of "global variable" feel from it. On the plus side, it is now importable and I can also use it in tests.

Suggestions?

rpmlb/builder/base.py
```diff
@@ -78,15 +90,40 @@ def before(self, work, **kwargs):
...
-        if 'macros' in package_dict:
-            self.edit_spec_file_by_macros(spec_file, package_dict['macros'])
-        if 'replaced_macros' in package_dict:
-            self.edit_spec_file_by_replaced_macros(
-                spec_file, package_dict['replaced_macros'])
+
+        spec_file_path = Path('{name}.spec'.format_map(package_dict))
+
+        with self.edit_spec_file(spec_file_path) as (source_file, target_file):
+            content_stream = iter(source_file)  # Start modifications
```
khardix: The `iter()` is not really needed here, but I need a new name for the iterator anyway (to allow for skipping of some edits), and I feel that `stream = iter(file)` is a bit better and more explicit than `stream = file`.

Counter-arguments welcome :)

tests/builder/test_base.py
```diff
@@ -125,3 +159,79 @@ def count_markers(path, marker='# Edited by rpmlb'):
         edit_file(empty_spec_path, indicator)
         assert empty_spec_path.exists()
         assert count_markers(empty_spec_path) == 1
+
+
+def test_add_macros(macro_spec_path, prepared_macros):
```
junaruga: Did you see my unit test's function name was reviewed by Miro, and those were changed like below lines.
Please change your function names "test_*" referring below lines.
https://github.com/sclorg/rpm-list-builder/blob/master/tests/test_recipe.py#L24-L66

tests/builder/test_base.py
```diff
@@ -202,7 +202,8 @@ def test_replace_macros(macro_spec_path, prepared_macros):
     result_macros = dict(MACRO_REGEX.findall(spec_contents))
 
     # All expected macros have their contents replaced
-    assert all(val == 'REPLACED' for val in result_macros.values()), result_macros
```
junaruga: The removed line was the line that you added on another commit on this PR.
I do not think this commit is meaningful for users that will see your commit from outside.
Please squash your commits with proper units. Please check other commit units in this PR.

junaruga: Perhaps, if you do not know how to squash and reabase for your commits?

This is the way.

```
git rebase -i a_commit_ref_before_your_1st_commit
```

tests/builder/test_base.py
```diff
@@ -125,3 +159,88 @@ def count_markers(path, marker='# Edited by rpmlb'):
         edit_file(empty_spec_path, indicator)
         assert empty_spec_path.exists()
         assert count_markers(empty_spec_path) == 1
+
+
+def test_macros_are_added_correctly(macro_spec_path, prepared_macros):
```
khardix: @junaruga 

>  Please change your function names "test_*" referring below lines.

I renamed the tests a bit more verbosely. Let me know if you would like more adjustments.

khardix:
@junaruga Commits were squashed, rebased and adjusted.

rpmlb/builder/base.py
junaruga: Why do you want to include "and escaped newlines"?
Matching `^%global\s+foo\s+var\s+$` is good enough?

khardix: Because formally, RPM macro bodies are allowed to contain white space and are allowed to span multiple lines, if the newlines are escaped: http://rpm.org/user_doc/macros.html

I wanted to make sure we do not left half of the body behind when replacing, should we ever need to replace a multi-line macro.

rpmlb/builder/base.py
```diff
@@ -126,31 +163,72 @@ def edit_spec_file(target_path: Path):
...
+            macros: Mapping of macro name to macro definition
+                    for all macros to add.
+
+        Yields:
+            Lines of the modified file.
+        """
+
+        # Prepend all definitions before the stream
+        for name, value in macros.items():
+            yield '%global {name} {value}\n'.format(name=name, value=value)
```
junaruga: You are using `yield` such as `return` in a part of `content_stream = self.add_macros`.
It is normal?

khardix: It is. `yield` turns the function into a generator. `content_stream` is that generator, which I subsequently manipulate.

rpmlb/builder/base.py
junaruga: ok. I agree that you want to cover exact macro specification.

junaruga:
I approve this code personally.
All the logic that was added in PR is passed the coverage test.

@hroncok any comments?

junaruga:
1st commit 30c06c3cba25ddf7eda023a80ff3e80342fca6e5 in this PR.
last commit 62f61b0693137f2f90cf9293c0fd567f122520a8 in this PR.

1st commit is not meaningful, isn't it?
Because last commit is fixing issue of 1st commit.

hroncok:
Will look at this this week.

khardix:
@junaruga 

> 1st commit is not meaningful, isn't it?
Because last commit is fixing issue of 1st commit.

Not really – 1st commit basically adds tests for `edit_spec_by_*` functions, so I know if I break something while rewriting it, which is done in the next two commits. From what I gathered, this is how TDD should look like.

The last commit redoes the `prepare` method, along with the test for it.

I guess if we would like to be extra consistent, I should either write all tests first and then all the implementation, or commit the implementation with the test for each modified method. I would prefer the former.

I will keep it in mind in the future; for now, I will wait for @hroncok's review and if needed, try to fix it with other requested changes (if any).

tests/builder/test_base.py
junaruga: @khardix yes I saw it. It's perfect for me.

junaruga:
About commit style, I still believe that the best solution is basically final status for commits is only 1 commit.
Because of my work experiment with popular Ruby packages.

I think that for example, you can find your preferred cases from below merged PR, I can believe your idea.
https://github.com/django/django/pulls?q=is%3Apr+is%3Aclosed

rpmlb/builder/base.py
hroncok: > (to allow for skipping of some edits)

I'm afraid I don't get it. (Sorry.) Why do you need a new name here?

tests/builder/test_base.py
```diff
@@ -14,12 +15,41 @@ def empty_spec_path(tmpdir):
...
         spec.unlink()
 
 
+@pytest.fixture
+def prepared_macros():
+    """Prepared testing spec macros"""
+
+    macro_dict = {
+        # Simple macro value
+        'test_macro_a': r'value',
```
hroncok: Why the raw strings here? To be consistent with test_macro_c?

tests/builder/test_base.py
```diff
@@ -125,3 +159,88 @@ def count_markers(path, marker='# Edited by rpmlb'):
...
+    # Count the macro occurrences
+    matches = MACRO_REGEX.finditer(spec_contents)
+    macro_counts = Counter(match.group('name') for match in matches)
+
+    # Each expected macro is present exactly once
+    assert all(cnt == 1 for cnt in macro_counts.values()), macro_counts
+
+    expected_macro_names = new_macro_dict.keys() | prepared_macros.keys()
+    present_macro_names = macro_counts.keys()
+    assert expected_macro_names == present_macro_names, present_macro_names
```
hroncok: I'm quite confident pytest would tell you why the `==` does not apply and what is the content of both `expected_macro_names` and `present_macro_names`, so adding it as the second assert argument is not necessary (but it doesn't harm) (applies to multiple locations)

hroncok:
In this case I'd either merge everything to one commit, or leave the last one separate.

hroncok:
I'v clicked on **Update branch** by accident. Please rebase before merging.

hroncok:
Rebased in https://github.com/sclorg/rpm-list-builder/pull/47

PR #38, opened 2017-06-20
Remove assert method that is used in a file under rpmlb.

junaruga:
This PR is for https://github.com/sclorg/rpm-list-builder/issues/23

> recipe.py
>   verify is bad.
>     assserts for logic
>     create a verify_attribute function?

I could not understand this part. "create a verify_attribute function".
Could you explain about this?

Thanks.

### DIFF
--- a/rpmlb/recipe.py
+++ b/rpmlb/recipe.py
@@ -61,6 +61,7 @@ class Recipe:
         if not isinstance(recipe['packages'], list):
             raise ValueError('packages should be a list.')
         for package_dict in self.each_normalized_package():
-            assert(package_dict['name'])
+            if not package_dict['name']:
+                raise ValueError('name is invalid in the package.')
 
         return True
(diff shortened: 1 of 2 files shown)


hroncok:
> I could not understand this part. "create a verify_attribute function".
> Could you explain about this?

(Actually not verify_attribute, but verify_key, sorry about that.)

Look at https://github.com/sclorg/rpm-list-builder/blob/v1.0.0/rpmlb/recipe.py#L54

The code is messy and can be replaced by a more general function (untested code follows)

```python
def verify_key(recipe, key, *, type_expected=object, optional=False):
    if key not in recipe:
        if optional:
            return
        raise ValueError('{} is required in the recipe.'.format(key))
    if not recipe[key] and not optional:
        raise ValueError('{} is invalid in the recipe, not true-ish.'.format(key))
    if not isinstance(recipe[key], type_expected):
        raise ValueError('{} is not a {}.'.format(key, type_expected))  # Maybe TypeError instead?

...

    def verify(self):
        verify_key(self.recipe, 'name')
        verify_key(self.recipe, 'packages', type_expected=list)
        verify_key(self.recipe, 'requires', type_expected=list, optional=True)

        for package_dict in self.each_normalized_package():
            assert(package_dict['name'])

        return True

```

hroncok:
Will merge.

PR #39, opened 2017-06-21
Add bootstrapping logic to recipe normalization

khardix:
Because Koji does not allow packages with the same NVR to be built multiple times, NVRs of repeated packages need to be adjusted when building them.

This change adds the necessary bookkeeping to the `Recipe` class.

For related discussion, see sclorg#32. Note that the implementation here differs from the suggested change.


rpmlb/recipe.py
```diff
@@ -23,22 +25,39 @@ def __init__(self, file_path, recipe_id):
...
+
+            package_dict = {
+                'name': name,
+                'bootstrap': bootstrap_position,
+            }
+
+            if isinstance(package, Mapping):
+                # package should be one-item dictionary
+                if len(package) != 1:
+                    msg = 'Invalid package data: {}'.format(package)
```
junaruga: I prefer "message" than "msg", Because as a variable to describe Message, "message" is used in this project.

```
$ grep -r msg -r rpmlb
$ grep -r message -r rpmlb
rpmlb/builder/base.py:            message = (
rpmlb/builder/base.py:            LOG.info(message)
rpmlb/builder/base.py:                message = 'pacakge_dict: {0}, num: {1}, work_dir: {2}'.format(
rpmlb/builder/base.py:                error = RuntimeError(message)
rpmlb/utils.py:        # Use absolute path to show it on FileNotFoundError message.
rpmlb/__init__.py:LOG_FORMAT = '%(asctime)s %(levelname)-8s %(name)-26s %(message)s'
rpmlb/downloader/base.py:            message = (
rpmlb/downloader/base.py:            LOG.info(message)
```

rpmlb/recipe.py
```diff
@@ -23,22 +25,39 @@ def __init__(self, file_path, recipe_id):
...
-                raise ValueError('package is invalid.')
+            name = self._package_name(package)
+            bootstrap_position = next(bootstrap_map[name], None)
+
+            package_dict = {
+                'name': name,
+                'bootstrap': bootstrap_position,
+            }
+
+            if isinstance(package, Mapping):
```
junaruga: This `isinstance(package, Mapping)` is better than `isinstance(package, dict)`?

rpmlb/recipe.py
```diff
@@ -64,3 +83,54 @@ def verify(self):
...
+        """
+
+        name_list = map(self._package_name, self.recipe['packages'])
+        count_map = Counter(name_list)
+
+        # For each package count it's instances except the last one;
+        # start at 1
+        sequence_map = {
+            name: iter(range(1, count))
+            for name, count in count_map.items()
```
junaruga: About your code.

```
name: iter(range(1, count))
for name, count in count_map.items()
```

Below 2 cases looks better than it?


```
name: iter(range(1, count))
    for name, count in count_map.items()
```

```
name: iter(range(1, count)) for name, count in count_map.items()
```

Or this code is including several statements per 1 line.
It is different from below document's policy.
Is it OK?

http://python-guide-pt-br.readthedocs.io/en/latest/writing/style/
> One statement per line

rpmlb/recipe.py
```diff
@@ -23,22 +25,39 @@ def __init__(self, file_path, recipe_id):
...
-                package_dict = package[name]
-                package_dict['name'] = name
-            else:
-                raise ValueError('package is invalid.')
+            name = self._package_name(package)
+            bootstrap_position = next(bootstrap_map[name], None)
+
+            package_dict = {
+                'name': name,
+                'bootstrap': bootstrap_position,
```
junaruga: Why do you want to define as value `1, 2, ...`? 

```
{
  'name': 'a'
  'bootstrap': '1'
},
{
  'name': 'a'
  'bootstrap': '2'
}
```

Below looks better isn't it?

```
{
  'name': 'a'
  'bootstrap': True
},
{
  'name': 'a'
  'bootstrap': True
}
```

And maybe you must write a document about the difference this `bootstrap` in this line, and `bootstrap` in a recipe file. Maybe you are progressing. Because I did not see the code for Builder class in this PR.

junaruga: Ah sorry I understand your idea the value: `1, 2 ...` is better than `True`, because we want to replace version number with the value.
If do that, I prefer `bootstrap_position` than `bootstrap`.
Because `bootstrap` looks flag for True/False, also `bootstrap` in recipe file and RPM spec file has been used as True/False flag. I want to distinguish from those.

rpmlb/recipe.py
khardix: `isinstance(package, dict)` forces the package to be an actual built-in dictionary or subclass.
`isinstance(package, Mapping)` says "anything that acts/says it acts as a dictionary", which is more in line with the python duck-typing way.

rpmlb/recipe.py
khardix: Good point, I will change the key to `bootstrap_position`.

rpmlb/recipe.py
khardix: I'm pretty sure that your second example does not fit within 79 characters :)

But the root cause here would probably be the "too many statements on one line" problem. I will try to come up with a better way to write this.

khardix:
Feedback accepted. I will push the fixes as a separate commits (so that we do not lose the comments on the file contents), which I will squash once the review is done.

khardix:
@junaruga The requested changes has been made.

Note that I have not rebased yet, so the py34 travis check will probably still fail.

rpmlb/recipe.py
junaruga: ok the current `Mapping` looks better than my suggestion.

junaruga:
@khardix great! I approved this PR personally.

hroncok:
Rebased in https://github.com/sclorg/rpm-list-builder/pull/46

PR #40, opened 2017-06-21
Allow type hinting in Python 3.4

khardix:
I would like to be able to specify type hints in function signatures. Unfortunately, the `typing` module, which is pretty much required for anything more complicated, is present in the standard library only from version 3.5.

I would like to add a `typing` dependency, so the type hinting could work even in the Python 3.4.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 PyYAML
 retrying
 click
+typing


hroncok:
We have dependencies in setup.py **and** requirements.txt. I think we need to add it in both places, possibly conditionalize it in setup.py using [the Environment Markers](https://www.python.org/dev/peps/pep-0496/).

PR #41, opened 2017-06-21
Fine-tune the typing installation

khardix:
Follow-up for #40, this PR adds the dependency to `setup.py` and also adds environment marker for it to install only on Py3.4 or older.


hroncok:
(Feel free to squash.  At least it would be clear why we made the travis changes.)

PR #49, opened 2017-06-28
Implement cmd element in a recipe file

junaruga:
This fixes https://github.com/sclorg/rpm-list-builder/issues/20

Just in case I also did below test with modified `ror.yml` recipe (See latest 1 commit): https://github.com/junaruga/rhscl-rebuild-recipes/commits/feature/ror-cmd-element .
 
https://github.com/sclorg/rpm-list-builder/wiki/Example-to-build-with-Copr-for-Ruby-on-Rails-recipe

```
$ rpmlb \
  --download custom \
  --build custom \
  --custom-file tests/fixtures/custom/rhpkg_internal_copr.yml \
  --work-directory $HOME/tmp/rpmlb_work_dir \
  ../rhscl-rebuild-recipes/ror.yml \
  rh-ror50 2>&1 | tee rpmlb.log
```

The `cmd` element is for a recipe file.
But I also modified `custom.py`  file for `custom` download/build mode.
Because I wanted to use the common logic to call commands in YAML file for from Recipe and Custom file logic.

### DIFF
--- a/rpmlb/custom.py
+++ b/rpmlb/custom.py
@@ -18,14 +18,15 @@ class Custom:
         if 'name' in kwargs:
             env['PKG'] = kwargs['name']
 
-        cmd_dict = self._get_yaml_content()
+        cmd_dict = self.yaml_content
         if key not in cmd_dict:
             return
         cmd_element = cmd_dict[key]
 
         Yaml.run_cmd_element(cmd_element, env=env)
 
-    def _get_yaml_content(self):
+    @property
+    def yaml_content(self):
         if self._yaml_content:
             return self._yaml_content
         yaml = Yaml(self._file_path)
--- a/rpmlb/custom.py
+++ b/rpmlb/custom.py
@@ -1,6 +1,5 @@
 import logging
 
-from rpmlb import utils
 from rpmlb.yaml import Yaml
 
 LOG = logging.getLogger(__name__)
@@ -19,20 +18,16 @@ class Custom:
         if 'name' in kwargs:
             env['PKG'] = kwargs['name']
 
-        for cmd in self.each_yaml_cmd(key):
-            utils.run_cmd(cmd, env=env)
-
-    def each_yaml_cmd(self, key):
-        cmd_dict = self._get_yaml_content(self._file_path)
+        cmd_dict = self._get_yaml_content()
         if key not in cmd_dict:
             return
-        cmds = cmd_dict[key]
-        for cmd in cmds:
-            yield cmd
+        cmd_element = cmd_dict[key]
 
-    def _get_yaml_content(self, file_path):
+        Yaml.run_cmd_element(cmd_element, env=env)
+
+    def _get_yaml_content(self):
         if self._yaml_content:
             return self._yaml_content
-        yaml = Yaml(file_path)
+        yaml = Yaml(self._file_path)
         self._yaml_content = yaml.content
         return self._yaml_content
--- a/rpmlb/yaml.py
+++ b/rpmlb/yaml.py
@@ -34,7 +34,6 @@ class Yaml:
 
     @staticmethod
     def run_cmd_element(cmd_element, **kwargs):
-        cmds = None
         if isinstance(cmd_element, str):
             cmds = [cmd_element]
         elif isinstance(cmd_element, list):
--- a/rpmlb/yaml.py
+++ b/rpmlb/yaml.py
@@ -3,6 +3,8 @@ import os
 
 import yaml
 
+from . import utils
+
 LOG = logging.getLogger(__name__)
 
 
@@ -29,3 +31,18 @@ class Yaml:
         output = yaml.dump(self.content, Dumper=Dumper,
                            default_flow_style=False)
         print(output)
+
+    @staticmethod
+    def run_cmd_element(cmd_element, **kwargs):
+        cmds = None
+        if isinstance(cmd_element, str):
+            cmds = [cmd_element]
+        elif isinstance(cmd_element, list):
+            cmds = cmd_element
+        else:
+            message = 'Invalid type of cmd_element: {!r}'.format(
+                type(cmd_element)
+            )
+            raise ValueError(message)
+        for cmd in cmds:
+            utils.run_cmd(cmd, **kwargs)
--- a/tests/test_custom.py
+++ b/tests/test_custom.py
@@ -1,4 +1,5 @@
 import os
+from unittest import mock
 
 import pytest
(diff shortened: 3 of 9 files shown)


tests/test_yaml.py
```diff
@@ -0,0 +1,93 @@
...
+    """A Yaml object with a valid YAML file."""
+    return Yaml(helper.get_valid_recipe_file())
+
+
+def test_init_loads_file():
+    yaml = Yaml(helper.get_valid_recipe_file())
+    assert yaml
+
+
+def test_init_load_default_loader_when_failing_to_import_c_loader():
```
junaruga: Below logic is for a case of getting ImportError, and run `from yaml import Loader` in `rpmlb/yaml.py`.
But I could not implement it. So, I appreciate if you give me some hints to implement that. 

```
    def __init__(self, file_path):
        try:
            from yaml import CLoader as Loader
        except ImportError:
            from yaml import Loader <= This logic!
```

junaruga:
@hroncok review, please! Thank you.

tests/test_yaml.py
hroncok: I don't understand what are you asking me to do.

junaruga: What I want to ask you is how to write a unit test for `rpmlb/yaml.py` the line: `from yaml import Loader` in the `__init__` function.

if you see `htmlcov/rpmlb_yaml_py.html` with your browser, you can see the line is not covered by the unit test.

hroncok: That's exactly why I don't like coverage hunting. You are chasing a ghost. No test for this is needed, unless you are trying to test if Python works.

junaruga: ok, I can understand what you said. I rebased!

junaruga:
Travis test is success in above Travis link. But in this page, the status is still not updated.

junaruga:
Rebased on laster master branch!

tests/test_yaml.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@pytest.fixture
+def valid_yaml():
+    """A Yaml object with a valid YAML file."""
+    return Yaml(helper.get_valid_recipe_file())
+
+
+def test_init_loads_file():
+    yaml = Yaml(helper.get_valid_recipe_file())
```
hroncok: There's a fixture with this exact content, why not use it?

tests/test_yaml.py
```diff
@@ -0,0 +1,77 @@
...
+    assert 'rh-ror50:' in stdout
+
+
+def test_run_cmds_runs_cmd_by_cmd_element_type_string():
+    # Test with generated file.
+    # Because we can not test using stdout with capsys fixture,
+    # the capsys does not capture stdout from subprocess in run_cmds.
+    random_file = None
+    try:
+        random_file = helper.get_random_generated_tmp_file()
```
hroncok: Can this fail with an exception? If not, i'd suppose you can move it out of the `try` block instead of assigning `None`. The same for the next test.

tests/test_yaml.py
```diff
@@ -0,0 +1,77 @@
...
+        env = {
+            'BAR': 'bar'
+        }
+        Yaml.run_cmd_element(cmd_element, env=env)
+
+        assert os.path.isfile(random_file_foo)
+        random_file_bar = random_file_bar_part + '-bar'
+        assert os.path.isfile(random_file_bar)
+    finally:
+        if os.path.isfile(random_file_foo):
```
hroncok: Create a function for this. Such as `remove_iff_isfile()` or `delete_file_lax()`, etc.

tests/test_yaml.py
junaruga: Because,
In case of that we implement like this.
If `valid_yaml` has error, I thought that it threw the error before `test_init_loads_file` was run. (I am not sure for that.)

```
def test_init_loads_file(valid_yaml):
    assert valid_yaml
```

junaruga:
Fixed for all the points that you mentioned, and rebased!

tests/test_yaml.py
hroncok: An error of a fixture is error of it's test. Please change it so it uses the fixture.

junaruga: ok, I changed. could you review again? Thanks.

tests/test_custom.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@pytest.fixture
+def valid_custom():
+    """A Custom object with a valid custom file."""
+    return Custom(helper.get_valid_custom_file())
+
+
+def test_init_loads_file():
+    custom = Custom(helper.get_valid_custom_file())
```
hroncok: Please, use the fixture, a discussed.

junaruga: ok, thanks for finding the issue of mine.

junaruga:
Rebased!

tests/test_custom.py
junaruga: @hroncok I rebased for the line. could you check again?

rpmlb/yaml.py
```diff
@@ -29,3 +31,18 @@ def dump(self):
         output = yaml.dump(self.content, Dumper=Dumper,
                            default_flow_style=False)
         print(output)
+
+    @staticmethod
+    def run_cmd_element(cmd_element, **kwargs):
+        cmds = None
```
khardix: I think this line is unnecessary – you do not need to declare a variable before `if` in python:

```python
>>> if True:
...     x = 42
...
>>> x
42
```

Unfortunately, I'm not able to find exact documentation on how this works, but it does…

rpmlb/custom.py
```diff
@@ -19,20 +18,16 @@ def run_cmds(self, key, **kwargs):
...
             return
-        cmds = cmd_dict[key]
-        for cmd in cmds:
-            yield cmd
+        cmd_element = cmd_dict[key]
+
+        Yaml.run_cmd_element(cmd_element, env=env)
 
-    def _get_yaml_content(self, file_path):
+    def _get_yaml_content(self):
```
khardix: This is asking for a `@property`:

```python
@property
def yaml_content(self):
    # The body stays the same
```

Usage: `cmd_dict = self.yaml_content`

Generally, any `get_anything(self)` method (getter with only `self` parameter) should be a `@property`.

rpmlb/yaml.py
junaruga: Wow, variable scope rule is unique for Python. OK I will fix it.

PR #50, opened 2017-06-30
Keep __init__.py as empty file as much as possible.

junaruga:
This PR is for below issue.
I moved the `configure_logging logic from `__init__.py` to `rpmlb/logging.py.

> https://github.com/sclorg/rpm-list-builder/issues/23
```
 __init__.py:
  nothing belongs here
```

@hroncok review please!


rpmlb/cli.py
```diff
@@ -1,24 +1,24 @@
...
+
 
 @click.command()
 # General options
 @click.option(
     '--verbose', '-v', is_flag=True, default=False,
     help='Turn on verbose logging.',
-    # Enable logging as early as possible
-    is_eager=True, expose_value=False,
-    callback=lambda ctx, param, verbose: configure_logging(verbose),
```
hroncok: Shouldn't this stay as callback?

tests/test_cli.py
```diff
@@ -83,21 +69,12 @@ def test_parse_argv_no_options(tmpdir, option):
...
-    assert LOG.getEffectiveLevel() == logging.INFO
-
-    recipe = Path('recipe.yml')
-    recipe.touch()
-
-    verbose_args = ['--verbose'] if verbose else []
-    level = logging.DEBUG if verbose else logging.INFO
-
-    runner.invoke(run, verbose_args + [str(recipe), 'test'])
-    assert LOG.getEffectiveLevel() == level
```
hroncok: You effectively stopped testing for what the test was testing before.

rpmlb/cli.py
junaruga: Yes, I think this callback should not stay here any more.
Because I moved and change a logic to initialize log settings in `__init__.py` to `rpmlb/logging.py`.
I want to call new `configure_logging` function as common logic that does not depend on `--verbose` option.

tests/test_cli.py
junaruga: Yes, I stopped the testing.

hroncok: Why?

junaruga: Because I thought the `test_log_verbosity` test was for below area. And I moved the callback logic from here to `run` function.
But if you like keeping the logic here. I will keep. I will rebase.

```
@click.option(
    '--verbose', '-v', is_flag=True, default=False,
    help='Turn on verbose logging.',
)
```

rpmlb/__init__.py
```diff
@@ -1,25 +1 @@
-"""Top-level module
-"""
-
-import logging
-
 __version__ = '1.0.0'
```
junaruga: I found below logic to manage version number. Does it look better than our current code?
https://github.com/rebase-helper/rebase-helper/blob/master/setup.py#L24

junaruga:
Rebased on latest master branch. and squashed.

hroncok:
Requesting @khardix to review this, because I'm not sure about that callback.

junaruga:
ok

rpmlb/__init__.py
khardix: If we really want to have completely empty `__init__.py`, than yes. However, I'm perfectly happy with leaving the version string here.

As a side note, what about [setuptools_scm](https://pypi.python.org/pypi/setuptools_scm)? It seems to be more convenient that tracking the version in a string somewhere…

tests/test_logging.py
```diff
@@ -0,0 +1,25 @@
...
+
+import pytest
+
+
+@pytest.fixture
+def rpmlb_logging():
+    from rpmlb import logging as rpmlb_logging
+    yield rpmlb_logging
+    # Reset logging module updated by logging.basicConfig.
+    importlib.reload(sys.modules['logging'])
```
khardix: Same as in `test_cli.py`.

rpmlb/__init__.py
junaruga: ok, thanks for the information. and welcome back :) I agree with you.

> As a side note, what about setuptools_scm? It seems to be more convenient that tracking the version in a string somewhere…

I am not sure about it.
So, if you like, I prefer you would send another PR for that after this PR.
Maybe @hroncok will accept it.

rpmlb/cli.py
junaruga: OK, could you send your PR to my branch: https://github.com/junaruga/rpm-list-builder/tree/feature/nit-picking-make-init.py-empty ?
Because I can not imagine your preferred implementation in this time.
I would accept the PR.

tests/test_cli.py
```diff
@@ -21,7 +22,7 @@ def runner():
             yield r
     finally:
         # Make sure other tests are not affected by the ones with --verbose
-        LOG.setLevel(logging.INFO)
+        importlib.reload(sys.modules['logging'])
```
junaruga: OK, could you send your PR to my branch: https://github.com/junaruga/rpm-list-builder/tree/feature/nit-picking-make-init.py-empty ?
Because I tried your way, and other ways. But still I can not pass unit tests on those ways.

PR #52, opened 2017-07-03
Set classifiers in setup.py.

junaruga:
It is as the title.
I did set seeing https://pypi.python.org/pypi?:action=list_classifiers

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,11 @@ setuptools.setup(
         'pytest-runner',
     ],
     classifiers=[
-        # TODO(Add items)
+        'License :: OSI Approved :: '
+        'GNU General Public License v2 or later (GPLv2+)',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.4',
+        'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
     ],
 )


setup.py
```diff
@@ -30,6 +30,13 @@
         'pytest-runner',
     ],
     classifiers=[
-        # TODO(Add items)
+        (
+            'License :: OSI Approved :: ',
+            'GNU General Public License v2 or later (GPLv2+)'
+        ),
```
hroncok: I know we are trying to please flake8, but this is uglier than a long line. The parenthesis are not necessary within a list definition.

hroncok: Also, with a comma, this is a tuple with two strs. I guess that was not intentional.

junaruga: Thanks for mentioning it. I tried to do this http://python-guide-pt-br.readthedocs.io/en/latest/writing/style/#line-continuations . But I mistook it.
Rebased!

setup.py
```diff
@@ -30,6 +30,11 @@
         'pytest-runner',
     ],
     classifiers=[
-        # TODO(Add items)
+        'License :: OSI Approved :: '
+        'GNU General Public License v2 or later (GPLv2+)'
```
hroncok: now there is a comma missing

junaruga: I am sorry. rebased with adding the comma.

junaruga:
> You might want to add Programming Language :: Python :: 3 :: Only, but it's not necessary.

Thanks for the reviewing.
Yes, maybe. Additional PR from you, is welcome!

PR #55, opened 2017-07-10
Add short help switch and better parameter names

khardix:
Fixes sclorg#53.

I added all the metavar names as they were before. The two exceptions are `--build` and `--download`, where the listing of accepted values seems more practical.


tests/test_cli.py
```diff
@@ -189,3 +189,15 @@ def test_simple_options(runner, recipe_arguments, option, value):
...
+
+@pytest.mark.parametrize('help_switch', ['--help', '-h'])
+def test_help_option_variants(runner, help_switch, recipe_arguments):
+    """Test that both short and long version of help switch works."""
+
+    # NOTE: recipe_arguments are necessary, both run.make_context and
+    # runner.invoke break without them for some reason
+    result = runner.invoke(run, [help_switch] + recipe_arguments)
+
+    print(result.output)
```
junaruga: This `print` is added intentionally?

khardix: Yes, for debugging purposes. If the tests fails, the help message is shown as captured stdout.

junaruga: OK.
You can see when the tests are success, the 2 help messages are printed on result of the unit tests.
Is this also showed intentionally?
https://travis-ci.org/sclorg/rpm-list-builder/jobs/251988985

khardix: Ah, apparently with `pytest -v` (verbose mode), the stdout is showed every time – which is unnecessary clutter. In this case, I will replace the print with something else.

hroncok: It should not work that way. That's actually weird.

khardix: @hroncok My bad, it is not the `-v` switch, it is the `-s` switch 😊

hroncok: And that's how it's supposed to work. Keep the print. Get rid of the -s switch if used.

khardix:
@junaruga I merged your PR 🎉 , can you approve the changes?

junaruga:
Yes, I approved! :)

Issue #48, opened 2017-06-28
platforms element in a recipe file


junaruga:
This ticket is to manage implementing `platforms` element
https://github.com/sclorg/rhscl-rebuild-recipes
`COLLECTION_ID/packages/PACKAGE/platforms`

## Format

The `platforms` file format is inspired from `.travis.yml`'s `branches` element.
https://docs.travis-ci.com/user/customizing-the-build/#Building-Specific-Branches

`COLLECTION_ID/packages/PACKAGE` is omitted in below examples.

```
# White list
platforms:
  only:
    - el7
    - el6
```
```
# Black list
platforms:
  except:
    - f
```

The platform string is such as `fc23`, `fc24`, `fc\d+`, `el6`, `el7`, `el\d+`.
(=> How about CentOS?)
White list can use regular expressions that is inspired from

> https://docs.travis-ci.com/user/customizing-the-build/#Building-Specific-Branches
> you can safelist them all with regular expressions, for example /^v\d+\.\d+(\.\d+)?(-\S*)?$/.

## Internal Logic

How to know current building platform that is used to build RPM package is Fedora, CentOS or RHEL from a RPM spec file's directory?

If it is in the RPM file, we can know using below macro.
But this way is not fit this time.

On fedora rawhide mock environment

```
<mock-chroot> sh-4.4# rpm -E '%{?fedora}'
27   
<mock-chroot> sh-4.4# rpm -E '%{?rhel}'
<= empty
<mock-chroot> sh-4.4# rpm -E '%{?fc27}'
1
<mock-chroot> sh-4.4# rpm -E '%{?fc26}'
<= empty
<mock-chroot> sh-4.4# rpm -E '%{?el7}'
<= empty
```

On RHEL7 mock environment
```
<mock-chroot> sh-4.2# rpm -E '%{?rhel}'
7
<mock-chroot> sh-4.2# rpm -E '%{?fedora}'
<= empty
<mock-chroot> sh-4.2# rpm -E '%{?el7}'
1
<mock-chroot> sh-4.2# rpm -E '%{?el6}'
<= empty
<mock-chroot> sh-4.2# rpm -E '%{?fc25}'
<= empty
```

junaruga:
Memo: After talking with @hhorak, I decided to add `rpmlb --platform` option.

junaruga:
@hhorak one more question, do you know CentOS's macro name for Fedora: `%{?fc25}`, RHEL: `%{el7}`?

[...]

PR #58, opened 2017-07-14
Add a KojiBuilder

khardix:
This is an implementation of a builder for Koji build service.

It should be usable with any Koji instance (i.e. cbs – CentOS Build Service or Fedora's Koji) for which exists a configuration profile (`/etc/koji.conf.d/*.conf`).


khardix:
The tests fails because of missing `rpmbuild` executable in the test environment. What options we have here?

junaruga:
There are several options.

1. Starting to use Docker for test on Fedora.
2. Mocking `rpmbuild` in unit test.
3. Installing `rpm` on Ubuntu.

But I prefer 1. using Docker.
Because we also need not only `rpmbulid` but also `mock` and etc in the future.
And it is exactly right environment for our test.


You can refer below example of mine for rpm Travis testing that I am trying to start to use it.
This example is run unit test on Docker without `Dockerfile`.
https://github.com/junaruga/rpm/blob/feature/starting-travis2/.travis.yml
https://github.com/junaruga/rpm/blob/feature/starting-travis2/scripts/run-in-container.sh
Build log (Failing right tnow)https://travis-ci.org/junaruga/rpm-list-builder/jobs/253554132

This is another example to use Docker with `Dockerfile`.
But preparing `Dockerfile` for each test cases is not good to maintain.
Maybe we can manage only 1 `Dockerfile`.
https://github.com/rpm-software-management/rpmlint/blob/master/.travis.yml

hroncok:
I also use this https://github.com/fedora-python/taskotron-python-versions/blob/develop/.travis.yml https://github.com/fedora-python/taskotron-python-versions/blob/develop/Dockerfile

khardix:
Well, since I'm untouched by Docker so far, any volunteers to set this up? 😄 

If it would take a long time, in the meantime I can mark the test as expected to fail when building in Travis.

junaruga:
> Well, since I'm untouched by Docker so far, any volunteers to set this up?

Yes, I can set up :) . I will send PR to your repository.

junaruga:
@hroncok In the case of tasktron with `Dockerfile`, the environment is 1 kind. If we can use common `Dockerfile` for several environments (platforms) such as Fedora rawhide, 25, 24, CentOS, that looks great.

khardix:
@junaruga Probably turn that into separate issue and PR altogether (see @hroncok's issue above and collaborate with him). After that is done, I will rebase this PR atop it.

junaruga:
@khardix ok!

rpmlb/builder/koji.py
```diff
@@ -0,0 +1,173 @@
...
+        """Initialize the slots"""
+
+        self.collection = None  #: The collection to build
+        self.epel = None  #: Version of EPEL to build for
+        self.profile = None  #: The koji configuration profile to use
+        #: The format string for construction of the actual template name
+        self.target_format = type(self).DEFAULT_TARGET_FORMAT
+
+    @property
+    def base_command(self) -> Sequence[str]:
```
junaruga: In case of creating sub class, there are 2 type of design for object oriented programming.
Those are "is-a" and "has-a"

Ref:
https://en.wikipedia.org/wiki/Is-a
https://en.wikipedia.org/wiki/Has-a

And Builder class's design that I intended, the relation of `BaseBuilder` and the sub classes is "is-a".

In the case of "is-a". `KojiBuilder` is a `BaseBuilder`.
That means that `KojiBuilder` should not expand a responsibility of `BaseBuilder`.
That means `KojiBuilder` specified functions such as `base_command` should be private method. `_base_command`.

rpmlb/builder/koji.py
```diff
@@ -0,0 +1,173 @@
...
+
+        binary = 'koji'
+
+        if self.profile is None:
+            return [binary]
+        else:
+            return [binary, '--profile', self.profile]
+
+    @property
+    def target(self) -> str:
```
junaruga: Why the public property is needed?

rpmlb/builder/koji.py
```diff
@@ -0,0 +1,173 @@
...
+        """The koji target to build into."""
+
+        return self.target_format.format(
+            collection=self.collection,
+            epel=self.epel,
+        )
+
+    def run(self, work: Work, **kwargs):
```
junaruga: Builder sub class should not override `run` function due to `is-a` design.
Builder class is also intended as a template method design pattern [1].
Super class defines the behavior with high abstraction level. and the sub classes only follow the behavior.

[1] https://en.wikipedia.org/wiki/Template_method_pattern

khardix: I agree, and this override is a currently necessary "hack".

The problem is that I need an access to `Work` object and the CLI arguments (`**kwargs`) to set the builder's parameters. The CLI parameters are passed to `build` method and I could set them there (which means setting them over for each package!), but I have no good way to access the `Work` object. The `prepare` cannot be used, as it is skipped on resume.

The probably cleanest way to remedy this would probably be extending the `__init__` method to have access to `Work` and the rest of the CLI arguments. However, this would mean changing all existing builders and the way the builder is loaded and constructed. That change is bit out of scope of just `KojiBuilder`…

Suggestions on how to do this better are welcome 😉. If nothing will pop up, I will prepare a new PR with the `__init__` changes suggested above.

rpmlb/builder/koji.py
hroncok: I see what you mean here, but I really see no benefit of taking class inheritance in Python so seriously. Where's the benefit?

khardix: @junaruga OK, I was not aware that in subclassing you might want to not extend the interface… but I see your point. I will turn this into `_base_command`.

hroncok: Also, expanding the "public" interface of a subclass goes against no rule in Is-a. Still you can just take an instance of KojiBuilder and assume it's a BaseBuilder - you can simply ignore the extra interface.

rpmlb/builder/koji.py
khardix: As above, it does not need to be "public", and I will make it "private". However, as it is dependent on other instance attributes, is should be dynamic property.

rpmlb/builder/koji.py
junaruga: Your question is why is following a typical design pattern such as template pattern beneficial?  
Hmm, the reason is same with why the template pattern exists.

It may be a little impolite for you.
Have you developed big application or service with a object oriented design?

hroncok: Depend's on what you call big. Nevertheless this smells like Java or C++ for me, that's why I'm asking.

junaruga: I respect you. Just I want to know your experience and background. There is a gap between us.

hroncok: Well you said it's because we are using `is-a`, not because we are using template method pattern. I just didn't understand why expanding a public interface of a subclass would violate an `is-a` relationship.

We may have different backgrounds, there might be a big gap between us, but I don't see why you are being so explicit about it instead of clearly explaining your intentions :confused:

hroncok: Keeping the "public" interface of all builder classes the same and making them all interchangable is good design. However an underscored method does not only indicate this is not part of the API, to me it also indicates "this is and implementation detail that may change at any time". Adding more "public" (= not underscored) methods is a promis for anyone who wants to subclass this that they can override those methods safely. I.e. I see two kinds of API here: one for a code that uses the builder class (and that one should remain interchengable) and one for any possible subclasses of this given builder. By adding extra methods to extend the second API you are not necessary breaking the first one.

junaruga: Just moment. Let me reply for your question.

hroncok: For example take a `RhpkgDownloader`. If `rhpkg` was not hardcoded but instead a "public" property, it would allow anyone to create a custom rhpkg based downloader with different command:

```python
class FedpkgDownloader(RhpkgDownloader):
    """Works like RhpkgDownloader but uses the fedpkg command instead"""
     command = 'fedpkg'
```

Now `BaseDownloader` does not have a "public" command property/attribute, but I see nothign wrong if `RhpkgDownloader` has it.

junaruga: > why you are being so explicit about it instead of clearly explaining your intentions

I am sorry. I was rude.

The class inheritance `is-a` is related to the template design pattern.

cli.py Sample.

```
# I intended factory method pattern in this line.
# People do not want to know how the builder instance was created,
# Also do not want to know which type of builder class and the detail of the instance.
#
# Factory method
# https://en.wikipedia.org/wiki/Factory_method_pattern#Python
#
# On only this line, people recognize giving build type as string.
# And after that, do not want to recognize sub class of the builder class.
builder = BaseBuilder.get_instance(A_BUILD_TYPE)

# People only want to know the "Builder" class (BaseBuilder)'s behavior (methods, properties)
builder.run()

# People can use the methods that are defined on BaseBuilder.py
# But can not access, use the methods/properties that are only defined on specified sub class of Builder. Using below methods is ok.
builder.before()
builder.after()
builder.prepare()

# But below case is not ok. Technically may be able to access in Python.
builder.a_sub_class_specified_method()

```

Another example.

```
builders = [
    BaseBuilder.get_instance(BUILD_TYPE_A)
    BaseBuilder.get_instance(BUILD_TYPE_B)
    BaseBuilder.get_instance(BUILD_TYPE_B)
]
  
for builder in builders:
    builder.run()
    somethings.
```

Another example.

```
builders = [
    BaseBuilder.get_instance(BUILD_TYPE_A)
    BaseBuilder.get_instance(BUILD_TYPE_B)
    BaseBuilder.get_instance(BUILD_TYPE_B)
]
  
for builder in builders:
    builder.run()
    somethings.
```

So, in this pattern, "a public interface of a subclass" (that can be `has-a` methods/properties) can not be accessed from outside of the class.

another example of Java

```
List list_a = new ArrayList()
List list_b = new Vector()

# as size() is defined in List interface, we can access.
list_a.size()
list_b.size()
```

> "one for any possible subclasses of this given builder"
The one (method) is called "protected" method, as you know.
The extend public interface(methods/properties) are only used from inherited class.

`_method_name`: protected method
`__method_name`: private method
https://stackoverflow.com/questions/11483366/protected-method-in-python

You do not like underscore method to describe the it is protected method?

”Polymorphism”
This same method name and different content is called Polymorphism.
https://en.wikipedia.org/wiki/Polymorphism_(computer_science)

"Loose coupling" is also a key idea.
We do not want to know many class from a class.
https://en.wikipedia.org/wiki/Loose_coupling

"Encapsulation" is also.
We only know only necessary things from outside.
https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)

rpmlb/builder/koji.py
junaruga: OK I still do not understand what you want.
Which class's object will access the `KojiBuilder`'s property `target`?

rpmlb/builder/koji.py
hroncok: There is no such thing as protected method in Python. There is only private with two underscores, but people are not using those (except for magic `__mehtods__` like this). One underscore in front of a method does not mean it's private nor protected, it does not mean anything. It's merely a convention. And explicit meaning of such underscore may vary form project to projet. As I read it (by experience) it usually means: _This is internal, it might change, it my go away, don't use it, don't override it in subclasses._

The stuff you talk about is mostly about languages where private and protected have meanings.

The Java example makes sense, you cannot access `list_a.some_method_a()` because you declared it a `List`. It has nothing to do with Python, because in python, you don't declare things as a specific type. Nevertheless nothing forbids you form creating such method and using it if the object was not declared as a `List`.

Look: a Vector in Java has a public method [capacity()](https://docs.oracle.com/javase/7/docs/api/java/util/Vector.html#capacity()) that is not specified in List interface or AbscractList base class. It's just there because *it is useful*.

Let me look at the Python example instead: Noone is calling a method only specified on KojiBuilder in a scope where it might be a different type of builder. The last line of the Python example you've provided is indeed wrong, no argue in that. But that doesn't mean in any way that simply presence of such method in the KojiBuilder class is wrong.

Polymorphism, Loose coupling, Encapsulation, nothing in there forbids you form extending the class interface in a subclass.

So let me repeat my question: **What is a benefit of not extending the API of a subclass?** Please be specific to our use case in KojiBuilder. Or, if you want to talk generally about OOP patterns, please show me where any of those terms you've used forbids to extend an API of a subclass.

rpmlb/builder/koji.py
junaruga: > However, as it is dependent on other instance attributes, is should be dynamic property.

Now I understand it. and I am fine for the dynamic property.
Yes, I think that "private" is better.

rpmlb/builder/koji.py
khardix: Clarification: I do not need the `Work` object itself, I need access to a collection ID, which I did not find elsewhere than `work._recipe._collection_id`.

junaruga: > However, this would mean changing all existing builders and the way the builder is loaded and constructed. That change is bit out of scope of just KojiBuilder…

I can understand what you want to do.
In the template design pattern, if we will follow the design pattern, it is better that `BaseBuilder` class `run` method is a template for all the builder builder sub classes.

Below is my suggestions for that.

> That change is bit out of scope of just KojiBuilder…

Yes, we have to change `BaseBuilder` for my suggestion. But I think that that's fine.

```
    def run(self, work, **kwargs):
        # init or another proper method name.
        init(work, kwargs)
        is_resume = kwargs.get('resume', False)
...

    def init(self, work, **kwargs):
        """Pre-run setup, not skipped on resume.
        Keyword arguments:
            work: Overview of the work to be done.
            koji_profile: The koji configuration profile to use.
            epel: The version of EPEL to build for.
        """
        pass
```

koji.py
```
    def init(self, work, **kwargs):
        self.collection = work._recipe._collection_id
        self.epel = kwargs['epel']
        self.profile = kwargs['koji_profile']
        self.target_format = ( 
            kwargs.get('koji_target_format')
            or type(self).DEFAULT_TARGET_FORMAT
        )
```

khardix: That is pretty much what I had in mind. Just please do not define extra `init` – lets use the the initializer `__init__` 😉.

junaruga: > Clarification: I do not need the Work object itself, I need access to a collection ID, which I did not find elsewhere than work._recipe._collection_id.

The argument `Work` may not needed by `KojiBuilder`.
Argument: collection_id or Argument: Work
We can think which is more natural as a template for possible builder sub classes.
I prefer using Argument: Work for `init` method in this case.

The "Work" object means working data or session data.
The merit of using Work object as argument is that people do not have to recognize which working object items (such as collection) is used in `init` method.

rpmlb/builder/koji.py
junaruga: > What is a benefit of not extending the API of a subclass?

The benefit is people can understand easily that it is intended as factory method pattern and template pattern. Though it 's hard to prove it is readable easily for everyone.

There is nothing to forbid it for these concepts.
Vector has a sub class specified method such as `capacity()` that is also used as argument of constructor, it is rare.

I agree to use a public unique method/property in a sub class.

I am sorry again.

khardix: Bit of a context from my side:

After in-person discussion about it, the fear that @junaruga had about this extending of subclass API was that it will confuse users. At the end we agreed that it is not an issue for us and decided to leave this properties "public" and leave the usage responsibility to user of the class.

@junaruga, feel free to add anything you deem necessary. From my point of view, this issue was addressed and resolved.

rpmlb/builder/koji.py
khardix: As above, at the end we agreed on leaving this "public".

junaruga: Yes, we agreed.

rpmlb/builder/koji.py
junaruga: @khardix the issue was resolved. I am fine.

hroncok: OK

khardix:
Rebased and adjusted as agreed. Ready for review once again, @junaruga, @hroncok.

junaruga:
It looks good to me!

rpmlb/cli.py
```diff
@@ -73,6 +73,22 @@
     '--copr-repo', '-C', metavar='COPR_REPO',
     help='Target Copr for copr builder.',
 )
+@click.option(
+    '--epel', metavar='VERSION', type=int, default=7,
```
junaruga: If `--epel` is only used for `--build koji`,  the name `--epel` confusing users?
`--koji-epel` is better?

rpmlb/cli.py
```diff
@@ -73,6 +73,22 @@
...
+@click.option(
+    '--koji-profile', metavar='PROFILE',
+    help='Configuration profile for koji builder.',
+)
+@click.option(
+    '--koji-target-format', metavar='FORMAT',
+    help='Target format string for koji builder.',
+)
+@click.option(
+    '--scratch', is_flag=True, default=False,
```
junaruga: If `--scratch` is only used for `--build koji`,  the name `--scratch` makes users imagine that the option is available for every builder? It will be common option for every builder in the future?

junaruga:
A just moment. I commented for option names.

rpmlb/cli.py
junaruga: Sorry my suggesting name `--koji-epel` was not good. 
How about a name `--rhel`?
Maybe it is used in RPM spec macro to describe what you want to use.

https://github.com/sclorg/rpm-list-builder/issues/48

```
%rhel 7
```

khardix: Right now, yes, `KojiBuilder` is the only builder that requires it. However, multiple builders might require specification of the OS version to use…

This might be also related to the [platform element](https://github.com/sclorg/rpm-list-builder/issues/48) in recipes – in effect, this is a very specialized platform choice switch.

Having said that, and since the platform element is in discussion, `koji-epel` would be indeed better for now. In the future, this information should be extracted from the target platform specification, however that would be specified.

TL;DR: I agree and will change to `koji-epel`. Once the platform specifiers are finished, this option should be replaced by them.

khardix: > Sorry my suggesting name `--koji-epel` was not good.

As per above comment – lets leave it at `koji-epel` for now, and replace it when the platform specifiers are implemented.

rpmlb/cli.py
khardix: Good point – not all builders would support scratch builds. To be consistent with others, I will use `koji-scratch`.

rpmlb/cli.py
junaruga: ok I agree with you.

rpmlb/cli.py
junaruga: ok, I agree with you.

junaruga:
OK, It good to me. I changed to Approved by myself in this case.

khardix:
@junaruga Thanks for review.

@hroncok Do you wan to add your review as well before I merge this?

hroncok:
> Do you wan to add your review as well before I merge this?

I don't think it is necessary, but if you explicitly want me to do it, I'll do it.

khardix:
> I don't think it is necessary, but if you explicitly want me to do it, I'll do it.

Not really, just making sure everyone is fine with this. Merging 🎉 🎆 🎉

PR #62, opened 2017-07-14
Use git tags as canonical versions

khardix:
See sclorg#60 for details.

Before (hardcoded):
```sh
$ python setup.py --version
1.0.0
```

After (generated from the latest git tag – `v1.0.0`):
```sh
$ python setup.py --version
1.0.1.dev59+gbe09bdd  # 59 commits since 1.0.0, git commit hash be09bdd
```


junaruga:
It looks good to me!

Issue #61, opened 2017-07-14
Run tests in Travis CI on Fedora with Docker


hroncok:
Related to https://github.com/sclorg/rpm-list-builder/pull/58

hroncok:
I'm working on this.

junaruga:
Oh you are working? OK. thanks.

PR #63, opened 2017-07-14
Travis CI: Run tests in Docker

hroncok:
Fixes https://github.com/sclorg/rpm-list-builder/issues/61

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+
 python:
 - '3.6'
 - '3.5'
@@ -6,23 +7,25 @@ python:
 
 matrix:
   include:
-    - python: 3.6
-      sudo: required
-      dist: trusty
+    - python: '3.6'
       env:
-      - TOXENV=style
-      before_install:
-      - sudo sed -i 's/trusty/yakkety/' /etc/apt/sources.list
-      - sudo rm -rf /etc/apt/sources.list.d/*
-      - sudo apt-get -qq update && sudo apt-get -qq install shellcheck
+      - TOXENV=style EXTRA_PKGS="ShellCheck which"
 
-cache:
-  directories:
-  - $HOME/.cache/pip
-  - .eggs
+sudo: required
+dist: trusty
+
+services:
+  - docker
 
 install:
-- pip install --upgrade pip setuptools
-- pip install tox-travis
+  - >
+    docker build --build-arg PYTHON=python${TRAVIS_PYTHON_VERSION}
+    --build-arg EXTRA_PKGS="$EXTRA_PKGS" -t rpmlb .
 
-script: tox
+before_script:
+  - sudo chown -R 1000 .
+
+script:
+  - >
+    docker run -v $(pwd):$(pwd) -w $(pwd) -i
+    -e TOXENV=${TOXENV:=py${TRAVIS_PYTHON_VERSION/./}} -t rpmlb
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ python:
 
 matrix:
   include:
+    - python: '3.4'
+      env:
+      - DNF=yum CONTAINER_IMAGE=jsmigel/centos-epel
     - python: '3.6'
       env:
       - TOXENV=style EXTRA_PKGS="ShellCheck which"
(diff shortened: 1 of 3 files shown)


hroncok:
Hmm, broken after https://github.com/sclorg/rpm-list-builder/pull/62

hroncok:
Installing git should fix it.

hroncok:
Fixed.

.travis.yml
```diff
@@ -1,28 +1,32 @@
...
-      dist: trusty
+    - python: '3.4'
       env:
-      - TOXENV=style
+      - DNF=yum OS=centos7
       before_install:
-      - sudo sed -i 's/trusty/yakkety/' /etc/apt/sources.list
-      - sudo rm -rf /etc/apt/sources.list.d/*
-      - sudo apt-get -qq update && sudo apt-get -qq install shellcheck
+      - sed -i 's@fedora@jsmigel/centos-epel@' Dockerfile
```
junaruga: The docker image `jsmigel/centos-epel` is better than other docker images `centos6` or `centos7`?
https://hub.docker.com/_/centos/

.travis.yml
```diff
@@ -1,28 +1,32 @@
...
+sudo: required
+dist: trusty
+
+services:
+  - docker
 
 install:
-- pip install --upgrade pip setuptools
-- pip install tox-travis
+  - docker build --build-arg PYTHON=python${TRAVIS_PYTHON_VERSION} --build-arg EXTRA="$EXTRA" --build-arg DNF=${DNF:=dnf} -t rpmlb .
```
junaruga: The line is too long. I prefer the line size is less than 78.
Can we use block style `>` [1]  like this? 

```
- >
  docker build
    --build-arg PYTHON=python${TRAVIS_PYTHON_VERSION}
    ..
```

[1] https://stackoverflow.com/questions/3790454/in-yaml-how-do-i-break-a-string-over-multiple-lines#21699210

junaruga: or shall we create a script file to run both `install` and `script` section?
Like this.
https://github.com/junaruga/rpm/blob/feature/starting-travis2/scripts/build.sh#L149-L165

.travis.yml
hroncok: We need epel. This basically adds epel over the centos image.

.travis.yml
```diff
@@ -1,28 +1,32 @@
...
+
+before_script:
+  - sudo chown -R 1000 .
 
-script: tox
+script:
+  - docker run -v $(pwd):$(pwd) -w $(pwd) -i -e TOXENV=${TOXENV:=py${TRAVIS_PYTHON_VERSION/./}} -t rpmlb
```
junaruga: Same with `install` section.

.travis.yml
hroncok: rebased

hroncok: > or shall we create a script file

IMHO two files are already more than enough

hroncok:
I screwed up the rebasing. will do again.

.travis.yml
junaruga: OK.
Another merit to create the script file is that people can try to run the docker test on local environment easily.
But in this PR, I agree with you.

.travis.yml
junaruga: ok, I understood it.

.travis.yml
```diff
@@ -1,28 +1,36 @@
...
 
 matrix:
   include:
-    - python: 3.6
-      sudo: required
-      dist: trusty
+    - python: '3.4'
       env:
-      - TOXENV=style
+      - DNF=yum OS=centos7
```
junaruga: Where is the environment variable `OS` used? I could not find the `OS` in a list of `Dockerfile`.
https://hub.docker.com/_/fedora/
https://hub.docker.com/r/jsmigel/centos-epel/

hroncok: It is not used, it's there only to be visible in the Travis build table, so the CentOS build will be easily distinguished when looking at it.

.travis.yml
hroncok: (We need epel for python34)

.travis.yml
junaruga: How about below way?
Defining not used variable for looking at it on Travis result is a little confusing me.

```
      env:
      - DNF=yum CONTAINER_IMAGE=jsmigel/centos-epel
      before_install:
      - sed -i "s@fedora@${CONTAINER_IMAGE}@" Dockerfile
```

.travis.yml
```diff
@@ -1,28 +1,36 @@
...
+      - sed -i 's@fedora@jsmigel/centos-epel@' Dockerfile
+    - python: '3.6'
+      env:
+      - TOXENV=style EXTRA="ShellCheck which"
 
-cache:
-  directories:
-  - $HOME/.cache/pip
-  - .eggs
+sudo: required
```
junaruga: We need 1 blank line before this line?

.travis.yml
junaruga: Sorry one more thing.
How about moving below line to common `before_install` section?
Because we can select docker image with $CONTAINER_IMAGE variable easily.

```
before_install:
  - if [ "${${CONTAINER_IMAGE}" != '' ]; then
        sed -i "s@fedora@${CONTAINER_IMAGE}@" Dockerfile
    if
```

.travis.yml
junaruga: ok, I did not know that.

.travis.yml
hroncok: No need, better readability.

PR #71, opened 2017-07-18
Refactor test helpers

khardix:
An implementation of sclorg#57.

General description of changes:
- Replaced custom setup/teardown logic with fixtures (built-in or custom) where possible
- Moved global helper functions to dedicated, auto-discovered namespace
- Replaced `str` with `pathlib.Path` for file-system path manipulation where possible


tests/conftest.py
```diff
@@ -88,3 +88,27 @@ def generate_tmp_path_list(count: int):
...
+
+@pytest.fixture
+def random_file_path():
+    """Provide path to a random temporary file."""
+
+    with generate_tmp_path() as path:
+        yield path
+
+
+@pytest.fixture
```
hroncok: Could this be session scoped? Paths are immutable.

tests/conftest.py
```diff
@@ -88,3 +88,27 @@ def generate_tmp_path_list(count: int):
...
+def valid_recipe_path():
+    """Provide path to a valid recipe file."""
+
+    return Path('tests/fixtures/recipes/ror.yml')
+
+
+@pytest.fixture
```
hroncok: Here as well.

khardix: It can be, in both places.

tests/test_recipe.py
```diff
@@ -3,13 +3,12 @@
...
 
-import helper
 from rpmlb.recipe import Recipe
 
 
 @pytest.fixture
-def ok_recipe():
-    return Recipe(helper.get_valid_recipe_file(), 'rh-ror50')
+def ok_recipe(valid_recipe_path):
+    return Recipe(str(valid_recipe_path), 'rh-ror50')
```
hroncok: Note to self: Would be good if Recipe could take a PathLib object.

khardix: It would, but since the `os.*` functions do not take PathLib object too (on Python 3.4), it is low priority issue…

PR #72, opened 2017-07-18
Reference RHSCL Rebuild Recipes

ncoghlan:
The RHSCL Rebuild Recipe README has some useful docs
for the recipe file format, so this adds a cross-reference from
the user guide.


hroncok:
Thanks.

PR #74, opened 2017-07-20
Koji hotfixes

khardix:
Fixes for two issues found when trying to rebuild `rh-ror50` with `KojiBuilder`:

* The `target_format` should be set to default value even when the `koji_target_format` is *explicitly* set to None
* `__init__` should accept (and ignore) arbitrary keyword arguments.

Tests triggering those two issues are included.

### DIFF
--- a/rpmlb/builder/koji.py
+++ b/rpmlb/builder/koji.py
@@ -180,7 +180,12 @@ class KojiBuilder(BaseBuilder):
 
         run_cmd_with_capture(' '.join(command))
 
-        srpm_path, = cwd.glob('{collection}-{name}-*.src.rpm'.format(
+        if name == collection:  # metapackage build
+            glob_format = '{collection}-*.src.rpm'
+        else:
+            glob_format = '{collection}-{name}-*.src.rpm'
+
+        srpm_path, = cwd.glob(glob_format.format(
             collection=collection,
             name=name,
         ))
--- a/rpmlb/builder/koji.py
+++ b/rpmlb/builder/koji.py
@@ -21,13 +21,16 @@ class KojiBuilder(BaseBuilder):
         koji_scratch: bool = False,
         koji_epel: int = None,
         koji_target_format: str = DEFAULT_TARGET_FORMAT,
-        koji_profile: Optional[str] = None
+        koji_profile: Optional[str] = None,
+        **extra_options
     ):
         """Initialize the builder.
 
         Keyword arguments:
             work: Overview of the work to do.
-            epel: Version number of the EPEL to build for.
+            koji_scratch: Indicates if the --scratch build option
+                should be used.
+            koji_epel: Version number of the EPEL to build for.
             koji_target_format: Format string for the target to build into.
             koji_profile: Name of the configuration profile to use.
         """
--- a/rpmlb/builder/koji.py
+++ b/rpmlb/builder/koji.py
@@ -1,5 +1,6 @@
 """Builder interface for the Koji build service."""
 
+import logging
 import re
 from pathlib import Path
 from typing import Iterator, Optional, Sequence
@@ -7,6 +8,8 @@ from typing import Iterator, Optional, Sequence
 from ..utils import run_cmd, run_cmd_with_capture
 from ..work import Work
 from .base import BaseBuilder
+
+LOG = logging.getLogger(__name__)
 
 
 class KojiBuilder(BaseBuilder):
@@ -21,6 +24,7 @@ class KojiBuilder(BaseBuilder):
         koji_scratch: bool = False,
         koji_epel: int = None,
         koji_target_format: str = None,
+        koji_owner: str = None,
         koji_profile: Optional[str] = None,
         **extra_options
     ):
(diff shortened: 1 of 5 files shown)


rpmlb/builder/koji.py
```diff
@@ -44,7 +44,10 @@ def __init__(
         self.epel = koji_epel
 
         #: Format string for the build target
-        self.target_format = koji_target_format
+        if koji_target_format is None:
+            self.target_format = type(self).DEFAULT_TARGET_FORMAT
```
hroncok: `self.DEFAULT_TARGET_FORMAT` does not work?

rpmlb/builder/koji.py
```diff
@@ -20,8 +20,9 @@ def __init__(
         *,
         koji_scratch: bool = False,
         koji_epel: int = None,
-        koji_target_format: str = DEFAULT_TARGET_FORMAT,
-        koji_profile: Optional[str] = None
+        koji_target_format: str = None,
+        koji_profile: Optional[str] = None,
+        **extra_options
```
junaruga: Where is `extra_options` used in the `__init__` method?

rpmlb/builder/koji.py
khardix: It should, but I was under the impression that you should not use `self` to access class-level attributes (as opposed to instance-level ones).

Unfortunately, I cannot remember where I found it or what was the reasoning behind that – so if you think it is superfluous, I will remove it.

rpmlb/builder/koji.py
khardix: Nowhere. It needs to be present in the signature, so we would not get "unexpected keyword argument" errors.

I know you were trying to get rig of unused parameters with some tool – if you know of a way to mark the argument as "unused, but it needs to be there", let me know.

rpmlb/builder/koji.py
hroncok: Never heard of that rule :confused:

hroncok: Oh. You should not assign to class level variable using self, because that would override it for the given object. But reading is fine.

hroncok:
```pycon
>>> class Foo:
...     clsvar = 1
...     def change(self):
...         self.clsvar = 2
... 
>>> Foo.clsvar
1
>>> foo = Foo()
>>> foo.clsvar
1
>>> foo.change()
>>> foo.clsvar
2
>>> Foo.clsvar
1
```

rpmlb/builder/koji.py
junaruga: Ah okay. no problem. :)

khardix:
@junaruga @hroncok Another round of fixes, this time related to SRPM naming and the necessity to add packages to a destination tag of a build target before the build can be submitted.

rpmlb/builder/koji.py
khardix: Ah, yes, this is probably what I heard and remembered it wrongly… I will get rid of the `type(…)` wrap.

khardix:
Test are failing because `scl-utils-build` is not present in the test environment, and thus the SCL-related macros are not defined. sclorg#84 should fix this.

khardix:
Note that the conflicts here (`Dockerfile`) are semi-deliberate – once the `rh-ror50` collection is successfully built and the changes are reviewed, I will be forced to make a rebase and come with nicer commit history 😄.

khardix:
Rebased onto current master.

junaruga:
@khardix 
I have a question.
There are 4 points using `is None`.
Those are below case? means you want to distinguish empty string/zero and None?


```
$ grep -r 'is None' rpmlb/builder/koji.py 
        if koji_epel is None:
        if koji_target_format is None:
        if self.profile is None:
        if position is None:
```

http://docs.python-guide.org/en/latest/writing/style/#check-if-variable-equals-a-constant
> # or, since None is considered false, explicitly check for it
> if attr is None:
>    print 'attr is None!'


Others look good to me.

khardix:
@junaruga In most of them, yes.

The first two (actually three, this PR add one more) are handling default arguments to `__init__` and need to know if the default should be used.

The `self.profile` one could be changed to `if not self.profile`, since empty profile name could be seen as no profile, but I would rather leave it as it stands (personal preference).

The `position` indicates position in bootstrap sequence, and position 0 != no position.

---

Note: I'm working on another change found during `rh-ror50` rebuild, so no need to review yet.

khardix:
I'm pleased to announce that after the last commit, the `rh-ror50` collection was completely [built in CBS](https://cbs.centos.org/koji/packages?tagID=1097).

That means this PR is no longer WIP, and ready for full review :tada:

@junaruga @hroncok Who will volunteer?

junaruga:
I am fine for current code.
But I want to ask to hear @hroncok 's opinion.

rpmlb/builder/koji.py
```diff
@@ -43,14 +50,23 @@ def __init__(
         self.epel = koji_epel
 
         #: Format string for the build target
-        self.target_format = koji_target_format
+        if koji_target_format is None:
+            self.target_format = self.DEFAULT_TARGET_FORMAT
+        else:
+            self.target_format = koji_target_format
```
hroncok: Is empty string a valid option? If not, this can be simplified to:

```python
self.target_format = koji_target_format or self.DEFAULT_TARGET_FORMAT
```

If empty string is a valid option, it can be changed to:

```python
self.target_format = koji_target_format if koji_target_format is not None else self.DEFAULT_TARGET_FORMAT
```

Which is a pretty long line, so maybe not that good.

khardix: Theoretically yes, but I doubt we will ever encounter configuration profile named `<empty string>`… so I will go with the first option.

The "ternary operator" was my first choice, but the line length was through the roof, so I have rewritten it to the `if` you see now.

khardix: Changed as requested.

khardix:
History rewritten to more logical chunks, and rebased on top of the current master. After Travis passes, I will merge this.

PR #77, opened 2017-07-22
Update .travis.yml and Dockerfile for readability.

junaruga:
I was inspired from https://github.com/rpm-software-management/rpm/blob/master/ci/Dockerfile
It looks better?


junaruga:
Thanks for the review.

PR #79, opened 2017-07-25
Set CUSTOM_DIR environment variable

### LINKED ISSUE #67: [RFE] Add CUSTOM_DIR environment variable
I've patched rpm-list-builder locally to set an additional `CUSTOM_DIR` variable when running custom download and build commands:

        env['CUSTOM_DIR'] = os.path.dirname(os.path.abspath(self._file_path))

This allows other files (such as mock configurations) to be located relative to the custom YAML file rather than relative to the package directory.

If such an addition were to be pushed back upstream, what else would be needed in order to include it?

ncoghlan:
This allows configuration files and helper scripts to be
located relative to the custom hook file that needs them.

For example, an uninstalled mock configuration file can be
referenced as `mock -r ${CUSTOM_DIR}/custom-mock.cfg ...`.

Closes #67

### DIFF
--- a/rpmlb/custom.py
+++ b/rpmlb/custom.py
@@ -1,4 +1,5 @@
 import logging
+import os.path
 
 from rpmlb.yaml import Yaml
 
@@ -9,7 +10,8 @@ class Custom:
     """A class to manage custom file."""
 
     def __init__(self, file_path):
-        self._file_path = file_path
+        self._file_path = file_path = os.path.abspath(str(file_path))
+        self._custom_dir = os.path.dirname(file_path)
         self._yaml_content = None
 
     def run_cmds(self, key, **kwargs):
@@ -23,6 +25,7 @@ class Custom:
             return
         cmd_element = cmd_dict[key]
 
+        env['CUSTOM_DIR'] = self._custom_dir
         Yaml.run_cmd_element(cmd_element, env=env)
 
     @property
--- a/tests/fixtures/custom/echo.yml
+++ b/tests/fixtures/custom/echo.yml
@@ -1,19 +1,26 @@
 before_download:
   - echo "This command is run before download."
-  - pwd
+  - |
+    echo "Using hook dir: ${CUSTOM_DIR}"
+    echo "Running hook in:" $(pwd)
 download:
   - echo "This command is run for each package download."
   - |
+    echo "Using hook dir: ${CUSTOM_DIR}"
+    echo "Running hook in:" $(pwd)
     echo "Downloading PKG: ${PKG}"
     if [ "${PKG}" != "" ]; then
         mkdir "${PKG}"
         touch "${PKG}/${PKG}.spec"
     fi
-  - pwd
 before_build:
   - echo "This command is run before build."
-  - pwd
+  - |
+    echo "Using hook dir: ${CUSTOM_DIR}"
+    echo "Running hook in:" $(pwd)
 build:
   - echo "This command is run for each package build."
-  - "echo Building PKG: ${PKG}"
-  - pwd
+  - |
+    echo "Using hook dir: ${CUSTOM_DIR}"
+    echo "Running hook in:" $(pwd)
+    echo "Building PKG: ${PKG}"
--- a/docs/users_guide.md
+++ b/docs/users_guide.md
@@ -196,7 +196,8 @@ work_directory/
   * `before_download`: Write commands to run before build.
   * `download`: Write commands to run for each packages in the pacakges directory. You can use environment variable `PKG` to describe the package name.
 
-
+For both hooks, the environment variable `CUSTOM_DIR` refers to the directory containing the custom file
+(allowing configuration files and helper scripts to be located relative to it).
 
 ### Specify work directory
 
@@ -258,6 +259,9 @@ work_directory/
 
   * `before_build`: Write commands to run before build.
   * `build`: Write commands to run for each packages in the pacakges directory. You can use environment variable `PKG` to describe the package name.
+
+For both hooks, the environment variable `CUSTOM_DIR` refers to the directory containing the custom file
+(allowing configuration files and helper scripts to be located relative to it).
 
 #### Don't build
 
--- a/tests/builder/test_custom.py
+++ b/tests/builder/test_custom.py
@@ -1,0 +1,36 @@
+from pathlib import Path
+from subprocess import STDOUT, check_output
+
+# These tests use subprocess directly due to weird interactions with test_cli
+# that lead to the command output being unreliable when using click.CliRunner.
+
+_TEST_DIR = Path(__file__).parent.parent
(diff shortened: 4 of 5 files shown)


ncoghlan:
@junaruga @hroncok I wasn't sure where or how to test this addition, so the PR doesn't currently include any new unit tests, it just adds a step to the custom echo file that prints out the new environment variable.

However, I'm wondering if it may make sense to consider moving the integration tests out of bash and into a framework like behave that can more easily capture the command output and make assertions about it.

junaruga:
@ncoghlan thank you for your pull-request. :)
Sorry right now I have a few tasks that I have to.
After that, I would take a look at this PR, and comment.

rpmlb/custom.py
```diff
@@ -23,6 +24,7 @@ def run_cmds(self, key, **kwargs):
             return
         cmd_element = cmd_dict[key]
 
+        env['CUSTOM_DIR'] = os.path.dirname(os.path.abspath(str(self._file_path)))
```
ncoghlan: The alternative way of fixing this would be `str(self._file_path.resolve().parent)` and dropping the `os.path` import.

However, I wasn't sure if `self._file_path` was guaranteed to always be a `pathlib.Path` instance, so coercing first and then using the `os.path` functions seemed more robust.

junaruga:
@ncoghlan Right now I checked your code and the document.
It looks awesome. Thanks.

Where we need to test is creating below new file.
Could you create the file, and add your test for your code?

```
tests/test_custom.py
```

Because the integration test is not good and it is deprecated. we decided to update test files for pytest instead of the integration test `run.sh`.

See https://github.com/sclorg/rpm-list-builder/issues/23
> I'd run integration tests using pytest as well.

```
tests/integration/run.sh
```

Thanks.

ncoghlan:
OK, I'll add a new test that runs the custom echo backend in a subprocess, and checks for the expected output.

Since it doesn't do any actual building, that should still be fast enough to be a reasonable unit test, and it will avoid having to design in any mocks.

I can also add an explicit test for #81 at the same time.

junaruga:
All right. Thank you.

ncoghlan:
Working on adding a test case for this, and it looks like there may be another test case which isn't reverting a monkeypatch correctly. My rationale for that is that when I do:

    python3 -m pytest -s tests -k custom_build_hooks

to run *just* the new test case, then I get the expected behaviour (i.e. I can see the output from running with the `echo.yml` configuration).

By contrast, running:

    python3 -m pytest -s tests -k custom

fails, and the output suggests that the custom YAML commands aren't actually being executed at all, so the build step fails due to the download step failing to create the required directories.

ncoghlan:
As a first guess, https://github.com/sclorg/rpm-list-builder/blob/master/tests/test_custom.py looks it may be the source of the problem, as it includes a couple of lines like:

    type(valid_custom).yaml_content = mock.PropertyMock(
                return_value=content,
    )

That skips reverting back to the original property descriptor after the test case is complete, which would explain the subsequent misbehaviour of testing a full custom build invocation.

I'll switch that over to using the `monkeypatch` fixture (https://docs.pytest.org/en/latest/monkeypatch.html) so that it gets reverted automatically after the test case finishes.

ncoghlan:
The above change fixed the custom backend invocation, but there's a separate problem in the interaction with `test_cli`, where that is somehow *turning off* the log capture for `runner.invoke`

khardix:
@ncoghlan `runner.invoke` does command-line parsing and callback invocation. There is a [callback][] that will [set the logging level][implementation] to `INFO` unless you pass the `--verbose` flag.

[callback]: https://github.com/sclorg/rpm-list-builder/blob/master/rpmlb/cli.py#L19
[implementation]: https://github.com/sclorg/rpm-list-builder/blob/master/rpmlb/logging.py#L15

ncoghlan:
I couldn't figure out what was going on with the failure to reliably capture output when using `click.CliRunner`, so I abandoned that approach and switched to using a simple subprocess invocation instead.

However, I'll be submitting a couple of other PRs that tidy up `test_custom` and `test_cli` a bit, in an effort to make them more reliably side-effect free.

ncoghlan:
@khardix Aye, and even the *normal* behaviour when running the test standalone was to only capture INFO messages and not the results of running the YAML hooks. The weird interaction was that the test would end up capturing *no* output, and instead print everything directly to the terminal.

Between click & pytest I decided there were just too many moving parts to try to debug, when using a real subprocess instead makes it straightforward to completely avoid any potential for interference.

ncoghlan:
@junaruga I've added a test that runs a custom "echo" build in a subprocess, and checks for the expected messages reporting the custom directory and the package names.

Using a full subprocess will also allow it to be extended to cover #81, but I figure that change belongs in a separate PR that also updates the documentation.

tests/fixtures/custom/echo.yml
```diff
@@ -1,19 +1,26 @@
...
         mkdir "${PKG}"
         touch "${PKG}/${PKG}.spec"
     fi
-  - pwd
 before_build:
   - echo "This command is run before build."
-  - pwd
+  - |
+    echo "Using hook dir: ${CUSTOM_DIR}"
+    echo "Running hook in:" `pwd`
```
junaruga: I would prefer

```
$(pwd)
```

than 

```
`pwd`
```
.

Because of https://github.com/koalaman/shellcheck/wiki/Sc2006

Could you change to it?
Other lines for \`pwd\` are also same.

tests/fixtures/recipes/echo-recipe.yml
```diff
@@ -0,0 +1,8 @@
+# Example recipe designed solely for use with the custom echo backend
```
junaruga: I would prefer the file name `echo_recipe.yml` than `echo-recipe.yml`.
Because custom files in `tests/fixtures/custom/` is using underscore for the delimiter.

tests/fixtures/custom/echo.yml
```diff
@@ -1,19 +1,26 @@
 before_download:
   - echo "This command is run before download."
-  - pwd
+  - |
+    echo "Using hook dir: ${CUSTOM_DIR}"
+    echo "Running hook in:" `pwd`
```
junaruga: There are 3 other points for \`pwd\`. Could you change those?

ncoghlan: Oops, fixed!

junaruga:
@ncoghlan Thanks!
So, after you squash 1 or 2 commit as you like, I would approve and merge it.

ncoghlan:
Done! Also 😃 at the typo in the last pre-squash commit message, as it didn't occur to me what this command would actually do:

    git commit -m "Replace other uses of `pwd`"

junaruga:
Thanks for the great work. Merged.

PR #80, opened 2017-07-25
Support distribution platform feature.

junaruga:
I implemented it.
Updating documents is an another PR after finishing this PR.

Specification: https://github.com/sclorg/rpm-list-builder/issues/48#issuecomment-315830205

Below `ror.yml` recipe file on feature/ror-dist-element branch is used for test 
The branch will be sent as pull-request after finishing this PR.
https://github.com/junaruga/rhscl-rebuild-recipes/commits/feature/ror-dist-element

### DIFF
--- a/rpmlb/cli.py
+++ b/rpmlb/cli.py
@@ -12,7 +12,7 @@ from .downloader.base import BaseDownloader
 from .recipe import Recipe
 from .work import Work
 
-DIST_RE = '^(fc|el|centos)\d*$'
+DIST_RE = re.compile('^(fc|el|centos)[0-9]*$')
 
 
 @click.command(add_help_option=False)
@@ -52,7 +52,7 @@ DIST_RE = '^(fc|el|centos)\d*$'
     help='Instructions for custom downloader and builder.',
 )
 @click.option(
-    '--dist', metavar='DIST',
+    '--dist', metavar=DIST_RE.pattern,
     help='Specifiy a distribution platform considered in a recipe file.',
     callback=lambda ctx, param, dist: validate_dist(dist),
 )
@@ -137,8 +137,7 @@ def run(recipe_file, collection_id, **option_dict):
 
 
 def validate_dist(dist):
-    pattern = re.compile(DIST_RE)
-    if dist and not pattern.match(dist):
+    if dist and not DIST_RE.match(dist):
         raise click.BadParameter(
-            'dist shoule be match to pattern {0}'.format(DIST_RE))
+            'dist shoule be match to pattern {0}'.format(DIST_RE.pattern))
     return dist
--- a/rpmlb/cli.py
+++ b/rpmlb/cli.py
@@ -2,6 +2,7 @@
 
 import logging
 import os
+import re
 
 import click
 
@@ -10,6 +11,8 @@ from .builder.base import BaseBuilder
 from .downloader.base import BaseDownloader
 from .recipe import Recipe
 from .work import Work
+
+DIST_RE = '^(fc|el|centos)\d*$'
 
 
 @click.command(add_help_option=False)
@@ -47,6 +50,11 @@ from .work import Work
     '--custom-file', '-c', metavar='CUSTOM_FILE',
     type=click.Path(exists=True, dir_okay=False, resolve_path=True),
     help='Instructions for custom downloader and builder.',
+)
+@click.option(
+    '--dist', metavar='DIST',
+    help='Specifiy a distribution platform considered in a recipe file.',
+    callback=lambda ctx, param, dist: validate_dist(dist),
 )
 # Download options
 @click.option(
@@ -126,3 +134,11 @@ def run(recipe_file, collection_id, **option_dict):
     builder.run(work, **option_dict)
 
     log.info('Success!')
+
+
+def validate_dist(dist):
+    pattern = re.compile(DIST_RE)
+    if dist and not pattern.match(dist):
+        raise click.BadParameter(
+            'dist shoule be match to pattern {0}'.format(DIST_RE))
+    return dist
--- a/rpmlb/downloader/base.py
+++ b/rpmlb/downloader/base.py
@@ -86,7 +86,6 @@ class BaseDownloader:
 
         dist = kwargs.get('dist')
         if dist and 'dist' in package_dict:
-            pattern = re.compile(package_dict['dist'])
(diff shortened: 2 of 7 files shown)


rpmlb/cli.py
```diff
@@ -48,6 +51,11 @@
     type=click.Path(exists=True, dir_okay=False, resolve_path=True),
     help='Instructions for custom downloader and builder.',
 )
+@click.option(
+    '--dist', metavar='DIST',
```
khardix: Suggestion: Replace the `DIST` metavar with the `DIST_RE` pattern (or similar), to inform user early what they may pass as a value:

```python
@click.option('--dist', metavar='(fc|el|centos)[0-9]*', ...)
```
```shell
$ rpmlb --help
...
--dist (fc|el|centos)[0-9]*    Specify a distribution...
...
```

rpmlb/cli.py
```diff
@@ -126,3 +134,11 @@ def run(recipe_file, collection_id, **option_dict):
     builder.run(work, **option_dict)
 
     log.info('Success!')
+
+
+def validate_dist(dist):
+    pattern = re.compile(DIST_RE)
```
khardix: `re.compile` within function seems pointless – you will compile the pattern *again* each time the function is called. Compile once, on the top of the file:
```python
DIST_RE = re.compile('^(fc|el|centos)\d*$')
```
and the use the result many times:
```python
if ... and not DIST_RE.match(dist):
    ...
```

And if you need to access the original text pattern:
```python
'dist should match pattern {0}'.format(DIST_RE.pattern)
```

tests/downloader/test_base.py
```diff
@@ -42,6 +47,49 @@ def test_run_does_not_call_before_build_after_with_resume():
...
 
 
+def test_run_skip_download(downloader):
+    downloader.download = mock.MagicMock(return_value=True)
+    downloader._is_download_skipped = lambda *args, **kwargs: True
+
+    mock_work = get_mock_work()
+    with pytest.helpers.generate_tmp_path() as tmp_path:
+        tmp_path_str = str(tmp_path)
+        os.mkdir(tmp_path_str)
```
khardix: You are casting `Path` to `str` where it is unnecessary. Better: `tmp_path.mkdir()`.

tests/downloader/test_base.py
```diff
@@ -42,6 +47,49 @@ def test_run_does_not_call_before_build_after_with_resume():
...
+        with pytest.helpers.pushd(tmp_path_str):
+            downloader.run(mock_work)
+            assert not downloader.download.called
+            assert os.path.isdir('a')
+            assert os.path.isdir('b')
```
khardix: Let's stick to `Path` for representing file-system paths:

```python
with pytest,helpers.pushd(tmp_path):  # the helper supports pathlib.Path
    ...
    assert Path('a').is_dir()
    assert Path('b').is_dir()
```
or
```python
    assert tmp_path.join('a').is_dir()
    assert tmp_path.join('b').is_dir()
```

tests/test_cli.py
```diff
@@ -152,6 +152,22 @@ def test_path_bad_permissions(path_kind, recipe_arguments):
                          options + recipe_arguments)
 
 
+@pytest.mark.parametrize('valid_dist',
+                         ('fc', 'fc26', 'el', 'el7', 'centos', 'centos7'))
+def test_dist_returns_value_on_valid_options(recipe_arguments, valid_dist):
+    options = ['--dist', valid_dist]
+    ctx = run.make_context('test_dist_returns_value_on_valid_options',
```
khardix: See sclorg#87 – please wrap the context with `with` statement:

```python
with run.make_context(...) as ctx:
# or
ctx = run.make_context(*long_argument_list)
with ctx:
```

Same on the test below.

rpmlb/cli.py
junaruga: The command line option `--dist`'s value is not regular expression.
Sorry. I changed the specification from last time I talked with you.

See https://github.com/sclorg/rpm-list-builder/issues/48#issuecomment-315830205 for detail.

Because if the value is regular expression `RE1`, it is impossible that a regular expression `RE1` compares with another regular expression `RE2` in recipe like below case.
I could not imagine the situation in advance.

```
if [[ RE1 =~ RE2 ]]; then
    something.
fi
```

Also user must know their executed distribution platform.
So, command line option `--dist` is just string.
So, `DIST` is good.

rpmlb/cli.py
junaruga: OK, thanks for mentioning it.
I will fix this.

tests/downloader/test_base.py
junaruga: OK. I will fix it.

tests/downloader/test_base.py
junaruga: OK I will fix this.

tests/test_cli.py
junaruga: OK, I will fix this.

rpmlb/cli.py
junaruga: Sorry I misunderstood your intention. You are right. I will change by your way.

junaruga:
@khardix I fixed and rebased on latest master branch.

rpmlb/downloader/base.py
```diff
@@ -58,3 +68,25 @@ def after(self, work, **kwargs):
...
+
+        if not package_dict:
+            raise ValueError('package_dict is required.')
+
+        is_skipped = False
+
+        dist = kwargs.get('dist')
+        if dist and 'dist' in package_dict:
+            pattern = re.compile(package_dict['dist'])
+            if not pattern.match(dist):
```
khardix: FYI: You do not need to compile the pattern for one-time use. This works just as well (and it it less clutter):

```python
if not re.match(package_dict['dist'], dist):
    ...
```

From [documentation](https://docs.python.org/3/library/re.html?#re.compile):

> The sequence
>
>     prog = re.compile(pattern)
>     result = prog.match(string)
>
> is equivalent to
> 
>     result = re.match(pattern, string)
>
> but using `re.compile()` and saving the resulting regular expression object for reuse is more efficient when the expression will be used several times in a single program.

junaruga: @khardix Thanks for mentioning it. I did not know the formula `re.match(pattern, string)`. So, I fixed with it.

PR #86, opened 2017-07-26
Revert monkeypatches in test_custom

ncoghlan:
Uses the pytest monkeypatch fixture to ensure that
changes to the Custom class are reverted after the
relevant tests.

### DIFF
--- a/rpmlb/yaml.py
+++ b/rpmlb/yaml.py
@@ -17,7 +17,7 @@ class Yaml:
         except ImportError:
             from yaml import Loader
         try:
-            with open(file_path, 'r') as stream:
+            with open(str(file_path), 'r') as stream:
                 self.content = yaml.load(stream, Loader=Loader)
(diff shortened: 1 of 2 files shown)


hroncok:
I think you've uncovered an issue, that the monkeypatching influenced other test and those test were not supposed to pass as they were.

https://github.com/ncoghlan/rpm-list-builder/blob/2dd627b6542d60dcaf1e838a4e1ec92e28cb2e98/rpmlb/yaml.py#L20 - this should also accept Path objects on older Pythons.

ncoghlan:
@hroncok Ah, that would be why I didn't see the failure locally - I hadn't done a full tox run since fixing this. Update pushed with the extra `str()` call.

hroncok:
Thanks. Now the Travis build pass. If you squash the first two commits, I'll merge.

ncoghlan:
@hroncok Rebased & squashed.

PR #87, opened 2017-07-26
Consistently close click contexts in test_cli

ncoghlan:
- use CliRunner.invoke wherever possible
- use with statements for other created contexts

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -113,12 +113,13 @@ def path_options(path_kind):
     return path, options
 
 
-def test_path_nonexistent(path_kind, recipe_arguments):
+def test_path_nonexistent(runner, path_kind, recipe_arguments):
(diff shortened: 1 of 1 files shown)


ncoghlan:
When I was attempting to use `CliRunner.invoke` in #79, I kept getting strange interactions with `test_cli` when I tried to run both of them in the same process.

While I never figured that out, and ended up switching to using a full subprocess for the #79 tests, this PR is a result of explore a theory that click may not like having multiple contexts created from a command without consistently closing them afterwards.

Given that it didn't actually seem to help resolve any of the side effects, I don't mind if this doesn't get merged, but also figured it made sense to offer it and let you make the decision on whether or not it is an improvement.

junaruga:
I merged it.
Thank you!

Issue #66, opened 2017-07-18
Support `fedpkg` in addition to `rhpkg`


ncoghlan:
Fedora's dist-git client is called `fedpkg` rather than `rhpkg`, which means some default rpm-list-builder features like the mock backend don't work as expected (it currently calls "rhpkg srpm" directly).

While it's easy enough to use a custom build to handle that case (which is what I'm currently doing), it probably makes more sense to make this configurable and/or based on automatic client detection somehow.

ncoghlan:
Note that the tests have example build configs for working with COPR and koji from a Fedora system: https://github.com/sclorg/rpm-list-builder/blob/master/tests/fixtures/custom/

So it may be possible to use a solution similar to the way `mock` handle buildroot configs, which is to have omitting the file extension indicate that the config should be looked up in a predefined configuration directory.

junaruga:
@ncoghlan Currently available download type is `none`, `local`, `rhpkg` and `custom` (default is `none`). The dist-git command `rhpkg` is used in Red Hat.
But as you said, `fedpkg` for Fedora should be supported as well.

Right now, the way to use both `fedpkg` and `mock` is

```
$ rpmlb -d custom -b mock ...
```

Your pull-request for the download type `fedpkg` option is welcome :)

> So it may be possible to use a solution similar to the way mock handle buildroot configs, which is to have omitting the file extension indicate that the config should be looked up in a predefined configuration directory.

I may not understand what you said.
There is a build type `mock`. You can use it with `--mock-config` (`-M`) option.

See https://github.com/sclorg/rpm-list-builder/blob/master/docs/users_guide.md

ncoghlan:
The specific problem is that `-b mock` calls `rhpkg srpm` to create the SRPM. As a result, it doesn't work on Fedora, which only has `fedpkg`.

However, it also turns out that there's a *separate* problem with combining that a `fedpkg srpm` call with custom downloads, which is that you also need to pass `--release el7` to ensure that `fedpkg` knows what release tag to use.

Currently I have that hardcoded in a custom build definition: https://github.com/ncoghlan/pyscl-devel/blob/master/rpmlb/sclorg-distgit-download.yml#L47

junaruga:
@ncoghlan I understood your situation. Sorry for that.
Right now the custom build looks the only way to build your packages.

Issue #83, opened 2017-07-25
Remove hardcoded rhpkg


junaruga:
## Issue

This issue is from

https://github.com/sclorg/rpm-list-builder/issues/66#issuecomment-316033852
> The specific problem is that -b mock calls rhpkg srpm to create the SRPM. As a result, it doesn't work on Fedora, which only has fedpkg.

Right now `rhpkg` is hardcoded.
It's time to fix it.

```
$ grep -r rhpkg rpmlb/
rpmlb/cli.py:    type=click.Choice('none local rhpkg custom'.split()),
rpmlb/cli.py:    help='Git branch for downloaders that use it (rhpkg).',
rpmlb/builder/mock.py:        utils.run_cmd('rhpkg srpm')
rpmlb/builder/copr.py:        utils.run_cmd('rhpkg srpm')
rpmlb/downloader/rhpkg.py:    """A downloader class to get a pacakge with rhpkg command."""
rpmlb/downloader/rhpkg.py:        self.do_rhpkg_and_checkout(package_dict, branch)
rpmlb/downloader/rhpkg.py:    def do_rhpkg_and_checkout(self, package_dict, branch):
rpmlb/downloader/rhpkg.py:        LOG.debug('rhpkg co %s', package)
rpmlb/downloader/rhpkg.py:        subprocess.check_call(['rhpkg', 'co', package])
```

## Possible solution

1. Add `--pkg-cmd [fedpkg|rhpkg]` option (default: fedpkg). The option name is inspired from `mock --dnf-cmd`.
2. Add `--rhpkg` option. If it is not specified, it is run as `fedpkg`.

But if `--download rhpkg` is specified, the package cmd is run as `rhpkg`.
But if `--download fedpkg` (future new feature) is specified, the package cmd is run as `fedpkg`.

Right now I prefer the solution 1.

hroncok:
What if we have something like BaseRpkgDownloader and inherit FedPkgDownloader and RhPkgDownloader from it?

```python
class FedPkgDownloader(BaseRpkgDownloader):
    command = 'fedpkg'
```

junaruga:
> What if we have something like BaseRpkgDownloader and inherit FedPkgDownloader and RhPkgDownloader from it?

That's great idea. I am going to use the way. Thanks.

hroncok:
Should be fixed by 9f74474044e34f835313420f1bfe3f8c6932ed2c https://github.com/sclorg/rpm-list-builder/pull/88

PR #88, opened 2017-07-27
Remove hardcoded "rhpkg".

junaruga:
This fixes https://github.com/sclorg/rpm-list-builder/issues/66 and https://github.com/sclorg/rpm-list-builder/issues/83 .

* Added --pkg-cmd PKG_CMD command option
  PKG_CMD works in build type: mock, copr.
* Added download type: fedpkg

Specification is here https://github.com/sclorg/rpm-list-builder/issues/83#issue-245364831 .

### DIFF
--- a/rpmlb/cli.py
+++ b/rpmlb/cli.py
@@ -29,7 +29,7 @@ DIST_RE = re.compile('^(fc|el|centos)[0-9]*$')
 )
 @click.option(
     '--download', '-d',
-    type=click.Choice('none local rhpkg custom'.split()),
+    type=click.Choice('none local fedpkg rhpkg custom'.split()),
     default='none',
     help='Choose a download type.',
 )
@@ -59,7 +59,8 @@ DIST_RE = re.compile('^(fc|el|centos)[0-9]*$')
 # Download options
 @click.option(
     '--branch', '-B', metavar='BRANCH',
-    help='Git branch for downloaders that use it (rhpkg).',
+    help=('Git branch for downloaders that use it '
+          '(download type: fedpkg, rhpkg).'),
 )
 @click.option(
     '--source-directory', '-S', metavar='SOURCE_DIRECTORY',
@@ -72,6 +73,12 @@ DIST_RE = re.compile('^(fc|el|centos)[0-9]*$')
     '--resume', '-r', metavar='RESUME',
     type=click.INT,
     help='Resume build from specified position.',
+)
+@click.option(
+    '--pkg-cmd',
+    type=click.Choice('fedpkg rhpkg'.split()),
+    default='fedpkg',
+    help='Choose a package command.',
 )
 @click.option(
     '--mock-config', '-M', metavar='MOCK_CONFIG',
@@ -110,6 +117,8 @@ def run(recipe_file, collection_id, **option_dict):
 
     log = logging.getLogger(__name__)
 
+    _adjust_option_dict(option_dict)
+
     # Load recipe and processing objects
     recipe = Recipe(recipe_file, collection_id)
     recipe.verify()
@@ -141,3 +150,15 @@ def validate_dist(dist):
         raise click.BadParameter(
             'dist shoule be match to pattern {0}'.format(DIST_RE.pattern))
     return dist
+
+
+def _adjust_option_dict(option_dict):
+    if not option_dict:
+        return
+    download = option_dict.get('download', None)
+    # When specifying download type: fedpkg or rhpkg,
+    # users must want to use the command in Builder.
+    for download_type in ('fedpkg', 'rhpkg'):
+        if download_type == download:
+            option_dict['pkg_cmd'] = download_type
+            break
--- a/rpmlb/cli.py
+++ b/rpmlb/cli.py
@@ -77,7 +77,6 @@ DIST_RE = re.compile('^(fc|el|centos)[0-9]*$')
 @click.option(
     '--pkg-cmd',
     type=click.Choice('fedpkg rhpkg'.split()),
-    default='fedpkg',
     help='Choose a package command.',
 )
 @click.option(
@@ -117,7 +116,7 @@ def run(recipe_file, collection_id, **option_dict):
 
     log = logging.getLogger(__name__)
 
-    _adjust_option_dict(option_dict)
+    _choose_pkg_cmd(option_dict)
 
     # Load recipe and processing objects
     recipe = Recipe(recipe_file, collection_id)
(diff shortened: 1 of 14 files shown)


junaruga:
I tested below cases too as acceptance test just in case.

```
$ rpmlb \
  --download fedpkg \
  --build mock \
  --mock-config fedora-rawhide-x86_64 \
  --branch master \
  --verbose \
  tests/fixtures/recipes/joke.yml \
  joke 2>&1 | tee rpmlb.log
```

```  
$ rpmlb \
  --download fedpkg \
  --build mock \
  --pkg-cmd fedpkg \
  --mock-config fedora-rawhide-x86_64 \
  --branch master \
  --verbose \
  tests/fixtures/recipes/joke.yml \
  joke 2>&1 | tee rpmlb.log
```

```
$ rpmlb \
  --download rhpkg \
  --build mock \
  --pkg-cmd rhpkg \
  --mock-config rhscl-2.4-rh-ror50-rhel-7-x86-64 \
  --branch rhscl-2.4-rh-ror50-rhel-7 \
  --verbose \
  tests/fixtures/recipes/ror.yml \
  rh-ror50 2>&1 | tee rpmlb.log
```

```
$ rpmlb \
  --download fedpkg \
  --build copr \
  --copr-repo rh-ror50-test \
  --branch master \
  --verbose \
  tests/fixtures/recipes/joke.yml \
  joke 2>&1 | tee rpmlb.log
```

```
$ rpmlb \
  --download rhpkg \
  --build mock \
  --mock-config rhscl-2.4-rh-ror50-rhel-7-x86-64 \
  --branch rhscl-2.4-rh-ror50-rhel-7 \
  --verbose \
  tests/fixtures/recipes/ror.yml \
  rh-ror50 2>&1 | tee rpmlb.log
```

tests/test_cli.py
```diff
@@ -9,7 +9,7 @@
 import pytest
 from click.testing import CliRunner
 
-from rpmlb.cli import run
+from rpmlb import cli
```
junaruga: This line is to use 2 methods in rpmlb.cli module in `test_cli.py` .

The way is from
http://docs.python-guide.org/en/latest/writing/structure/
> Best
> x = modu.sqrt(4)  # sqrt is visibly part of modu's namespace

rpmlb/downloader/base_rpkg.py
```diff
@@ -0,0 +1,29 @@
...
+from .. import utils
+from .base import BaseDownloader
+
+LOG = logging.getLogger(__name__)
+
+
+class BaseRpkgDownloader(BaseDownloader):
+    """A downloader class to get a pacakge with package command."""
+
+    def command(self):
```
khardix: This should be a `@property`.

I would envision the whole inheritance tree to look something like this:

```python
class Base:
    @property
    def command(self):
        raise NotImplementedError('...')

    def echo_command(self):
         """Example method that uses the command property."""
        print(self.command)


class A(Base):
    command = 'rhpkg'


class B(Base):
    command = 'fedpkg'
```
```python
>>> a, b = A(), B()
>>> a.echo_command()
rhpkg
>>> b.echo_command()
fedpkg
```

rpmlb/downloader/fedpkg.py
```diff
@@ -0,0 +1,6 @@
+from .base_rpkg import BaseRpkgDownloader
+
+
+class FedpkgDownloader(BaseRpkgDownloader):
+    def command(self):
```
khardix: See comment for `BaseRpkgDownloader.command`.

rpmlb/downloader/rhpkg.py
```diff
@@ -1,35 +1,6 @@
...
-
-        LOG.debug('rhpkg co %s', package)
-        subprocess.check_call(['rhpkg', 'co', package])
-        LOG.debug('cd %s', package)
-        os.chdir(package)
-        LOG.debug('git checkout %s', branch)
-        subprocess.check_call(['git', 'checkout', branch])
+class RhpkgDownloader(BaseRpkgDownloader):
+    def command(self):
+        return 'rhpkg'
```
khardix: See comment for `BaseRpkgDownloader.command`.

rpmlb/cli.py
```diff
@@ -126,3 +135,15 @@ def run(recipe_file, collection_id, **option_dict):
     builder.run(work, **option_dict)
 
     log.info('Success!')
+
+
+def _adjust_option_dict(option_dict):
```
khardix: This is a very generic name. Lets call it `_choose_pkg_cmd` or something like that.

rpmlb/cli.py
```diff
@@ -126,3 +135,15 @@ def run(recipe_file, collection_id, **option_dict):
...
+    if not option_dict:
+        return
+    download = option_dict.get('download', None)
+    # When specifying download type: fedpkg or rhpkg,
+    # users must want to use the command in Builder.
```
khardix: That is not true. What about rebuilding from Fedora to (EP)EL/CentOS? In that case you might want a `fedpkg` download, but `rhpkg` build.

Better solution would be setting the `pkg_cmd` default to `None` and do something like this:

```python
# If the user did not request a pkg command,
# choose one based on relevant download type.

# user chose the command explicitly
if option_dict['pkg_cmd'] is not None:
    return

# no explicit command, but relevant builder was selected
if download in ('fedpkg', 'rhpkg'):
    option_dict['pkg_cmd'] = download
    return

# no pkg_cmd or relevant builder – set default value
option_dict['pkg_cmd'] = 'fedpkg'
```

At that point, the functions starts to look as a prime candidate for `pkg_cmd` callback, although that might be more complicated to use.

rpmlb/downloader/base_rpkg.py
hroncok: Definitively agreed (commenting, so you get e-mail).

junaruga: I also agree :)

junaruga: I could not have the idea last time.

rpmlb/downloader/fedpkg.py
junaruga: Yes, ok!

rpmlb/downloader/rhpkg.py
junaruga: Yes, ok!

rpmlb/cli.py
junaruga: ok, the concrete name makes sense. I will fix it.

rpmlb/cli.py
junaruga: OK, it looks better. I will fix it.

Maybe below looks better.
Ref: http://docs.python-guide.org/en/latest/writing/style/#check-if-variable-equals-a-constant

```
# user chose the command explicitly
if not option_dict['pkg_cmd']:
    return
```

khardix: I tend to be explicit with `None`, as for me it signifies "unknown or unset value" in opposite to "set value that happens to be falsey", (i.e. empty string). But this boils down to personal preference, so I'm fine with your proposed solution too.

junaruga: Ah okay I could understand your intention.
However the comment "user chose the command explicitly" helps users to understand your intention.
So, I will keep this code.
Thanks for explanation.

junaruga:
Merged with squash.
The document is later.

Issue #90, opened 2017-07-28
README not on PyPI


hroncok:
In order to display README's content on PyPI, we need:

 * [ ] convert README.md to README.rst
 * [ ] add the file's content to long_description

hroncok:
On it.

junaruga:
Thank you 😃 
Now we have pypi page.
https://pypi.python.org/pypi/rpmlb/

PR #91, opened 2017-08-01
Convert README to rst, use it in long_description

hroncok:
Fixes https://github.com/sclorg/rpm-list-builder/issues/90


README.rst
```diff
@@ -0,0 +1,172 @@
...
+
+    $ which rpmlb
+    /usr/local/python-3.6.1/bin/rpmlb
+
+    $ rpmlb -h
+
+Usage
+-----
+
+See `Users Guide <https://github.com/sclorg/rpm-list-builder/blob/master/docs/users_guide.md>`_.
```
hroncok: Changed for full URL, so the link works on PyPI as well.

README.rst
```diff
@@ -0,0 +1,172 @@
...
+
+Run below command to check installed command.
+
+::
+
+    $ source venv/bin/activate
+    (venv) $ venv/bin/rpmlb -h
+    (venv) $ deactivate
+
+Install directly by pip
```
khardix: Since we are/will be on PyPI, I would move this sub-section to the top of the install section and simplify it to `pip3 install rpmlb`.

junaruga: I agree with your opinion.
But I think this PR's purpose is to convert from markdown to rst file to show it on pypi.
So, your thoughts can be another PR.

hroncok: Agree with @junaruga. Also it's hard to track changes made during format conversion.

khardix: OK, then I'm fine with this ☺.

Issue #15, opened 2017-05-03
Add CLI option to select number of retries


khardix:
Right now, the number of retries for failed build is hard-coded to 3. During the debugging of a recipe, disabling the retrying would be beneficial.

I would like to have the option to set the number of retries from the command line.

junaruga:
I like it too.

ncoghlan:
For sclo-python, I'd definitely like the ability to disable this behaviour - I'd prefer for an unreliable build to just fail outright, rather than for it to give the appearance of working even though some package builds aren't working first time every time.

junaruga:
The option: default: 0 (disable retry) may be good, considering your opinion.

PR #92, opened 2017-08-03
Add CLI option to select number of retries

khardix:
This is implementation of issue #15. As discussed there, the default number of retries is now 0 (in other words, only one attempt is made for a build before reporting the failure).


rpmlb/cli.py
```diff
@@ -75,6 +75,11 @@
     help='Resume build from specified position.',
 )
 @click.option(
+    '--retry', metavar='N',
```
junaruga: Referring current other options' metavar rule, I prefer `RETRY` or `RETRY_NUM` than `N`.

rpmlb/cli.py
```diff
@@ -75,6 +75,11 @@
...
+    type=click.IntRange(min=0), default=0,  # no negatives
+    help='Retry failed build N times.',
```
junaruga: I want you to include the word "Retry number" in the help message.
And `N` can be changed to another metaver name.

rpmlb/builder/base.py
```diff
@@ -80,7 +80,13 @@ def run(self, work, **kwargs):
                     continue
 
                 self.prepare(package_dict, **kwargs)
-                self.build_with_retrying(package_dict, **kwargs)
+
+                # Repeat the call in case of failure
+                retry_call(
+                    self.build, fargs=(package_dict,), fkwargs=kwargs,
+                    tries=kwargs.get('retry', 0)+1,  # 0 tries == no call
```
junaruga: I like below kind of way, because of 1 line 1 statement rule.

```
retry_num = kwargs.get('retry', 0)
retry_call(
    self.build, fargs=(package_dict,), fkwargs=kwargs,
    tries=retry_num+1
...
```

tests/builder/test_base.py
```diff
@@ -122,11 +122,7 @@ def test_run_exception(builder):
     if sys.version_info >= (3, 6):
         builder.build.assert_called()
     # bulid should be called 3 times for retry setting.
-    calls = [
-        mock.call({'name': 'a'}),
-        mock.call({'name': 'a'}),
-        mock.call({'name': 'a'}),
-    ]
+    calls = [mock.call({'name': 'a'}, retry=3)] * 3
```
junaruga: I think adding below cases can be better. This line shows called 3 times. but should be called 4 times.

If `builder.run(mock_work, retry=3)`, => builder.build should be called 4 times (not 3 times, not 2 times.).
If `builder.run(mock_work, retry=0)`, => builder.build should be called 1 time (not 2 times).

(Sorry my previous retry specification was my mistake. I called `builder.build` total 3 times for retry=3).

But interestingly `builder.build.assert_has_calls(calls)` is passed for current 3 times calls.
So, maybe `assert_has_calls` can not test how many times the method is called with same argument.

rpmlb/cli.py
khardix: What about `RETRY_COUNT`? That I can ever write into the help message nicely 😉.

rpmlb/builder/base.py
khardix: Yeah, my bad 😞. I will rewrite that.

tests/builder/test_base.py
khardix: > If `builder.run(mock_work, retry=3)`, => builder.build should be called 4 times (not 3 times, not 2 times.).
> If `builder.run(mock_work, retry=0)`, => builder.build should be called 1 time (not 2 times).
> ...
> But interestingly builder.build.assert_has_calls(calls) is passed for current 3 times calls.

That is interesting and I missed that. I will look into it.

rpmlb/cli.py
junaruga: OK I like `RETRY_COUNT`. :)

tests/builder/test_base.py
khardix: `assert_has_calls` does a *subset* of the calls – which means if there any extra calls, the assert still passes.

I fixed that by also asserting on `Mock.call_count`.

khardix:
@junaruga Fixed as discussed.

junaruga:
It looks good to me.
Thanks for investigating `builder.build.call_count`.
I would approve it.
Please squash 2 commits to 1 commit when merging.

PR #95, opened 2017-08-08
Fix for pylint unused-argument, unused-variable warnings.

junaruga:
I ran `pylint` to know Python better practices in my repository (https://github.com/junaruga/rpm-list-builder/tree/feature/add-pylint) as a trial.

And I fixed below unused-argument and unused-variable warnings.
How do you think?

```
rpmlb/builder/custom.py:29: [W0613(unused-argument), CustomBuilder.before] Unused argument 'kwargs'
rpmlb/builder/mock.py:29: [W0613(unused-argument), MockBuilder.before] Unused argument 'kwargs'
rpmlb/builder/base.py:31: [W0613(unused-argument), BaseBuilder.__init__] Unused argument 'work'
rpmlb/builder/base.py:258: [W0613(unused-argument), BaseBuilder.prepare_extra_steps] Unused argument 'package_metadata'
rpmlb/builder/koji.py:21: [W0613(unused-argument), KojiBuilder.__init__] Unused argument 'extra_options'
rpmlb/downloader/base.py:48: [W0612(unused-variable), BaseDownloader.run] Unused variable 'num_name'
```

### DIFF
--- a/rpmlb/downloader/base.py
+++ b/rpmlb/downloader/base.py
@@ -45,7 +45,7 @@ class BaseDownloader:
 
         self.before(work, **kwargs)
 
-        for package_dict, num_name in work.each_num_dir():
+        for package_dict, _ in work.each_num_dir():
             if self._is_download_skipped(package_dict, **kwargs):
                 name = package_dict['name']
                 LOG.debug('Skip download package: %s', name)
--- a/rpmlb/builder/base.py
+++ b/rpmlb/builder/base.py
@@ -28,7 +28,7 @@ MACRO_REGEX = re.compile(
 class BaseBuilder:
     """A base class for the package builder."""
 
-    def __init__(self, work: Work, **options):
+    def __init__(self, _: Work, **options):
         """Initialize the builder to a valid state.
 
         Keyword arguments:
@@ -255,7 +255,7 @@ class BaseBuilder:
     def prepare_extra_steps(
         self,
         source: Iterator[str],
-        package_metadata: Mapping[str, Any]
+        _: Mapping[str, Any]
     ):
         """Builder-specific package preparation.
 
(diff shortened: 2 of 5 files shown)


hroncok:
I don't think it adds readability much. Actually I think it makes the code less readable.

khardix:
> I don't think it adds readability much. Actually I think it makes the code less readable.

Agreed. If we really want this kind of checking, we need a way to mark an argument as "deliberately unused" (i. e. part of the API callback). AFAIK `pylint` does not provide something like that out of the box, so we would need to come with our own **documented** convention.

junaruga:
I agree with you, guys.

There are several cases for the warnings in current source code.

Which case is better to use underscore variable?

I want to investigate those cases more when I have a time.
For example, what is a better practice to use a type hints and a kwargs together? (Case 3)
So, I want to keep this ticket right now.

* Case 1. rpmlb/builder/base.py:
  * `__init__`: Related to get_instance
  * `prepare_extra_steps`: Related to child class's structure.
* Case 2. rpmlb/builder/custom.py, rpmlb/builder/mock.py
  * Related to parent class.
* Case 3. rpmlb/builder/koji.py
  * Related to combination with type hints and kwargs.
* Case 4. rpmlb/downloader/base.py
  * Just unused variable for `work.each_num_dir`'s return values.

rpmlb/downloader/base.py
```diff
@@ -45,7 +45,7 @@ def run(self, work, **kwargs):
 
         self.before(work, **kwargs)
 
-        for package_dict, num_name in work.each_num_dir():
+        for package_dict, _ in work.each_num_dir():
```
junaruga: This case is a good one to use underscore variable, isn't it?

> https://stackoverflow.com/questions/5893163/what-is-the-purpose-of-the-single-underscore-variable-in-python
> As a general purpose "throwaway" variable name to indicate that part of a function result is being deliberately ignored,

PR #105, opened 2018-02-21
Fix naming of packages added to target koji tag

khardix:
`KojiBuilder` was incorrectly handling naming of metapackages when adding them to target tag, essentialy duplicated the collection name (i.e. adding `rh-ror50-rh-ror50` in place of just `rh-ror50`).

This fix adds a check for the metapackage name, avoiding the duplication.

### DIFF
--- a/rpmlb/builder/koji.py
+++ b/rpmlb/builder/koji.py
@@ -131,51 +131,41 @@ class KojiBuilder(BaseBuilder):
         """Build a package using Koji instance"""
 
         base_name = package_dict['name']
-        full_name = self._full_package_name(base_name, self.collection)
-        srpm_path = self._make_srpm(
-            name=base_name,
-            collection=self.collection,
-            epel=self.epel,
-        )
+        full_name = self._full_package_name(base_name)
+        srpm_path = self._make_srpm(name=base_name)
 
         self._add_package(full_name)
         self._submit_build(srpm_path)
         self._wait_for_repo()
 
-    @staticmethod
-    def _full_package_name(base_name: str, collection: str) -> str:
+    def _full_package_name(self, base_name: str) -> str:
         """Determine what the full name of the package should be
         after applying the collection prefix.
 
         Keyword arguments:
             base_name: Name of the package, without the collection part.
-            collection: Name/identification of the package's collection.
 
         Returns:
             Canonical full name.
         """
 
         # The metapackage
-        if base_name == collection:
+        if base_name == self.collection:
             return base_name
         else:
-            return '-'.join((collection, base_name))
+            return '-'.join((self.collection, base_name))
 
-    @staticmethod
-    def _make_srpm(base_name: str, collection: str, epel: int) -> Path:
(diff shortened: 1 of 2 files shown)


rpmlb/builder/koji.py
```diff
@@ -136,10 +136,12 @@ def build(self, package_dict, **kwargs):
             epel=self.epel,
         )
 
-        self._add_package(name='{collection}-{name}'.format(
-            collection=self.collection,
-            name=package_dict['name'],
-        ))
+        if package_dict['name'] == self.collection:  # metapackage
```
hroncok: Make this a function. You can then test the function isolated without all the bloody mocking.

khardix: Yeah, bit of a brainf*rt on my part. I will take time to do this properly next week.

Thanks for keeping the technical debt at bay ☺ !

khardix:
Next attempt at this fix. This time, I left it in two commits:

1. The first one (c8ff0ca9bfd253d5ef5fba6b0d187bc62ba5e1bb) contains the fix itself, extracted into separate function (as requested), with test and associated changes in related functions.
2. The other one is refactoring (making regular methods from static ones, extracting repeated test setup into fixture, etc.).

rpmlb/builder/koji.py
```diff
@@ -130,33 +130,52 @@ def prepare_extra_steps(self, source, package_dict):
...
+        Keyword arguments:
+            base_name: Name of the package, without the collection part.
+            collection: Name/identification of the package's collection.
+
+        Returns:
+            Canonical full name.
+        """
+
+        # The metapackage, or the base_name already has the prefix applied.
+        if base_name.startswith(collection):
```
hroncok: Should it not start with `colection+'-'`?

khardix: No, because the name of the metapackage is exactly `collection`. `base_name` is not NVR, just the "N" part.

hroncok: So in fact it should be: `if base_name.startswith(collection + '-') or base_name == collection:`?

khardix: Basically, yes. The important part is the `base_name == collection`. The `startswith` check is best-effort to avoid duplicate attachment of the prefix if the name already starts with it (and so in fact it is `full_name`).

If we are afraid of false positives here, I would drop the `startswith` check entirely.

hroncok: I'm just being paranoid. However I was bit by almost the same thing in the past. See https://github.com/pytest-dev/pytest/issues/2939#issuecomment-346207679

khardix: I understand. The trouble here is that I'm not sure if the `'-'` at the end would help anyhow 😉.

Since the problem I'm trying to solve here manifests (so far) only on metapackages, I will drop the `startswith` check, which should make the condition unambiguous. It can be always added back if we run into a case that needs it.

hroncok: Do it. Thank you.

Issue #106, opened 2018-04-15
Style error for lint_bash.sh


junaruga:
https://travis-ci.org/sclorg/rpm-list-builder/jobs/366491730

```
pushd "$(dirname "${0}")/.." > /dev/null
^-- SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
```

SC2164
https://github.com/koalaman/shellcheck/wiki/SC2164

SC2039 defined in lint_bash.sh
https://github.com/koalaman/shellcheck/wiki/SC2039

PR #107, opened 2018-04-15
Fix style check issue for ShellCheck.

junaruga:
This fixes https://github.com/sclorg/rpm-list-builder/issues/106 .

The reason is ShellCheck changed the definition of the check?
In my memory, `disable=SC2039` was to restrict the use of `pushd` and `popd`.
Anyway, I fixed, and tested the outdated `tests/integration/run.sh` script too.

One test case was failed because of the error of "fedpkg srpm" for rh-ror50.
Maybe this test script does not follow the feature to add `fedpkg`.

```
$ python3 -m venv ./venv

$ source venv/bin/activate

(venv) $ pip install -e .

(venv) $ which rpmlb
~/git/sclorg/rpm-list-builder/venv/bin/rpmlb

(venv) $ sh tests/integration/run.sh 
== TEST: download_only_test ==
  > [OK]

== TEST: download_from_local_and_build_with_dummy_test ==
  > [OK]

== TEST: download_by_rhpkg_and_build_with_dummy_test ==
  > [OK]

== TEST: download_with_custom_and_build_with_dummy_test ==
  > [OK]

== TEST: build_only_verbose_test ==
  > [OK]

== TEST: build_only_with_mock_test ==
  > [FAILURE]

== TEST: build_only_with_custom_echo_test ==
  > [OK]

== TEST: build_only_with_custom_mock_test ==
  > [OK]

== TEST: build_only_resume_test ==
  > [OK]

== TEST: download_and_build_with_custom_echo_test ==
  > [OK]

== TEST: download_and_build_with_custom_mock_test ==
  > [OK]


real	7m29.410s
user	2m50.111s
sys	1m22.038s
exist status: 1

$ vi integration.log
...
2018-04-15 13:38:10,623 ERROR    rpmlb.utils                CMD: [fedpkg srpm] failed at [/home/jaruga/git/sclorg/rpm-list-builder/tmp/work_directory/1/rh-ror50]
...
```


junaruga:
Sorry I created the branch to 'origin/hotfix/fix-bash-style` by my mistake. I should create it to my repository.

scripts/lint_bash.sh
```diff
@@ -1,9 +1,8 @@
 #!/bin/bash -v
-# shellcheck disable=SC2039
 
-pushd "$(dirname "${0}")/.." > /dev/null
+pushd "$(dirname "${0}")/.." > /dev/null || exit 1
 ROOT_DIR=$(pwd)
-popd > /dev/null
+popd > /dev/null || exit 1
```
hroncok: I'd rather call the script in a fashion that makes it fail if anything fails (`bash -eu`), rather than having it full with this nonsense. If you have to fight the linter, maybe it's a bad linter :(

junaruga: I removed `|| exit 1`, added `set -e` in the head of each file, the linter was passed. and ran the script. `lint_bash.sh` works. `integration/run.sh` does not work. I do not want to fix the issue of `integration/run.sh`, because it is outdated script.

Below is other possibilities to fix. Which do you like or any other suggestions?
SC2164 is the target lint check code for this time.

1. `scripts/lint_bash.sh`
  a. Add  `# shellcheck disable=SC2164` to the top of the file.
  b. Add `set -e` to the top of the file.
2. `tests/integration/run.sh`.
  b. Add  `# shellcheck disable=SC2164` to the top of the file.

or

3. Stop using `ShellCheck`.

hroncok: I'd rather stop using `ShellCheck`, but that's up to you.

junaruga: @hroncok I see, you would prefer the "3. Stop using ShellCheck".
@khardix how do you think?

junaruga: I changed my mind. I like to stop using `ShellCheck`. I do not want to fight for the linter continuously.
If it would be python package, I could run specific version's `ShellCheck` easily.

junaruga:
I removed `ShellCheck` and the related code. Could you check and merge?
Thanks.

junaruga:
Thanks for the merging and deleting the remote branch.

Issue #108, opened 2018-05-13
Using rpm-list-builder for Python 3.7 rebase in Fedora


hroncok:
I'm slowly rebuilding all our Python packages in Copr: [@python/python3.7](https://copr.fedorainfracloud.org/coprs/g/python/python3.7/builds/)

I want to prep everything and once it's working, use the [yaml file](https://github.com/hroncok/rpm-list-builder/blob/python37/python37.yaml) to rebuild stuff in Fedora. As I'm looking at rpm-list-builder, I'm not sure I can use it out of the box, so I'm opening this issue to track things.

What I need to do when I finally run this in Fedora proper.

 1. Use fedpkg as source :heavy_check_mark: 
 2. When rebuilding, bump the release and changelog entry
    * This should be possible with a custom build command and `rpmdev-bumpsec && git commit && git push && fedpkg build` right?
 3. Before rebuilding next package, use `koji wait-repo` to wait for the previous package to be available in the buildroot
    * This should be appendable to the custom build command
    * I'd need to get the NVR to wait for, this should do: `rpm --define 'dist .fc29' -q --qf "%{NAME}-%{VERSION}-%{RELEASE}\n" --specfile *.spec | head -n1`
 4. When I build something without custom macros, I don't want to have the `# Edited by rpmlb` commited. Is it possible to disable this edit?
 5.  When I build something with custom macros and later without, I want the second bump to revert the changes (see example bellow). This is the part when it gets tricky I guess.

```yaml
    - python-pluggy:
        macros:
          _without_tests: 1
    - pytest
    - python-pluggy
```

This should first commit the following pseudopatch to python-pluggy:

```patch
+# Edited by rpmlb
+%global _without_tests 1
 ...
-Release: 1
+Release: 2
 ...
 %changelog
+* Sun May 13 2018 Me Myself <me@fedoraproject.org> - 1.0-2
+- Bootstrapping Python 3.7
+
```

And later commit this:

```patch
-# Edited by rpmlb
-%global _without_tests 1
 ...
-Release: 2
+Release: 3
 ...
 %changelog
+* Sun May 13 2018 Me Myself <me@fedoraproject.org> - 1.0-3
+- Bootstrapping Python 3.7
+
 * Sun May 13 2018 Me Myself <me@fedoraproject.org> - 1.0-2
 - Bootstrapping Python 3.7
```

This is something that seems almost impossible currently.

My idea is:

- when doing macro changes, save the diff (prior to running bumpspec)
[...]

PR #110, opened 2018-06-04
Remove the marker to edit.

junaruga:
Related https://github.com/sclorg/rpm-list-builder/issues/108#issuecomment-388814996


junaruga:
@hroncok thanks for the checking!

Issue #111, opened 2018-08-13
Removing or moving internal documents to github


junaruga:
As you know, we have internal documents. I think it's time to remove the documents.
It has not been used long time.

Some of the documents might be moved to this repository.
But I think mentioning https://pypi.org/project/rpmrh/ as a variant of this tool is good enough.

How do you think?

hroncok:
I don't know about any documents.

junaruga:
@hroncok I sent you the URL on email.

hroncok:
I agree that mentioning https://pypi.org/project/rpmrh/ as a variant of this tool is good enough. Everything else is not useful.

junaruga:
@hroncok all right. @khardix how about you?

khardix:
I'm fine with that.

PR #112, opened 2018-08-14
Add a document for a variant software of RPM List Builder.

junaruga:
This fixes https://github.com/sclorg/rpm-list-builder/issues/111 .
I removed the internal documents.
This PR is to add a document to explain a variant of RPM List Builder.

Here is the added document.
https://github.com/junaruga/rpm-list-builder/blob/feature/doc-variant/docs/variant.md


junaruga:
Okay, thanks for the checking.

PR #113, opened 2019-04-25
Retrieve sources from lookaside cache with {fed,rh}pkg

khardix:
When using the `{fed,rh}pkg` downloader, also retrieve the sources from the lookaside cache.

---

The koji builder cannot build an SRPM in the working directory if it does not have all the package sources and patches available, since it does not rely on any `*pkg` command, but uses `rpm`/`rpmbuild` directly. The other relevant builders (except the custom one, which leaves that up to the user) all use `*pkg srpm` for building the SRPM, which calls `*pkg sources` if necessary.

The koji builder could of course be adapted to using the `*pkg` command internally, but I believe this to be an unnecessary coupling/dependency. With this patch, both ways will be working.

Thoughts?

### DIFF
--- a/rpmlb/downloader/base_rpkg.py
+++ b/rpmlb/downloader/base_rpkg.py
@@ -22,8 +22,13 @@ class BaseRpkgDownloader(BaseDownloader):
 
         package = package_dict['name']
         cmd = r'''
-{} co {} && \
-cd {} && \
-git checkout {}
-        '''.strip().format(self.command, package, package, branch)
+{command} co {package} && \
+cd {package} && \
+git checkout {branch} && \
+{command} sources
+        '''.strip().format(
+            command=self.command,
+            package=package,
+            branch=branch,
+        )
         utils.run_cmd(cmd)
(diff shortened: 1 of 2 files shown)


hroncok (approved):
So the downloader is now handling the downloads? Works for me.

Also, if you use custom downloaded, the non-custom fedpkg builder will fetch the sources for you via `fedpkg srpm` anyway, so we are not breaking anything. Also works for me.

hroncok:
(Assuming the test will pass which they don't.)

rpmlb/downloader/base_rpkg.py
```diff
@@ -22,8 +22,9 @@ def download(self, package_dict, **kwargs):
...
         cmd = r'''
-{} co {} && \
-cd {} && \
-git checkout {}
-        '''.strip().format(self.command, package, package, branch)
+{command} co {package} && \
+cd {package} && \
+git checkout {branch} && \
+{command} sources
+        '''.strip().format(command=self.command, package=package, branch=branch)
```
hroncok: > E501 line too long (80 > 79 characters)

khardix: OK, I will wrap that line.

That reminds me that we should probably add a project flake8 configuration, since mine ignore line length warnings (compatibility with [black](https://github.com/ambv/black)) – so this test passes on my machine :)

junaruga:
Thanks for merging!

